PR #4, opened 2017-03-31
Let's be more verbose about production-readiness, etc.

burke:


README.md
```diff
@@ -47,3 +47,21 @@ Bootsnap.setup(
 ```
 
 **Protip:** You can replace `require 'bootsnap'` with `BootLib::Require.from_gem('bootsnap', 'bootsnap')` using [this trick](https://github.com/Shopify/bootsnap/wiki/Bootlib::Require). This will help optimize boot time.
+
+### How likely is this to work?
+
+We use the `*_path_cache` features in production and haven't experienced any issues in a long time.
+
+The `compile_cache_*` features work well for us in development on macOS, but probably don't work on
+linux at all.
```
jules2689: `linux` -> `Linux`

Issue #17, opened 2017-04-28
Minimum Ruby version support?


bf4:
Bootsnap seems to contain >= Ruby 2.3-specific code. I haven't found where this is mentioned.  First failure on Ruby 2.2.6 is from the `<~EOF` heredoc.  I suspect subsequent failures regarding finding libyaml are similar since switching to Ruby 2.3.3 resulted in expected behavior.

burke:
https://github.com/Shopify/bootsnap/pull/18

It's not just cosmetic; the API to marshal ruby bytecode was introduced in 2.3. We've added a minimum required version to the gemfile now though.

PR #18, opened 2017-04-28
Lock down ruby version

shiroyasha:
Solves https://github.com/Shopify/bootsnap/issues/17.

Please review 🙇

### DIFF
--- a/bootsnap.gemspec
+++ b/bootsnap.gemspec
@@ -23,6 +23,8 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
   spec.extensions    = ['ext/bootsnap/extconf.rb']
 
+  spec.required_ruby_version = '>= 2.3.0'
+
   spec.add_development_dependency "bundler", '~> 1'
   spec.add_development_dependency 'rake', '~> 10.0'
   spec.add_development_dependency 'rake-compiler', '~> 0'


jules2689:
If the changes are minimal to support more ruby, it might be beneficial to do so. More people using this means more community driven support and development. Wdyt @burke?

@shiroyasha @bf4 do you happen to know what pieces are 2.3.0 specific? I saw bf4 mention the EOF part.

jules2689:
@shiroyasha mind adding an extended commit message explaining why 2.3.0 was chosen, then LGTM

shiroyasha:
@jules2689 sure 👍 

Do you prefer messages on pull-requests, or on the actual commits?

jules2689:
@shiroyasha I think in this case I'd like the context to be available when someone clones this locally. I thought about asking for an inline comment as well, but didn't like the idea of leaking implementation details about something that could change into the code itself, so a commit seemed like a good compromise with these in mind.

shiroyasha:
@jules2689 updated the commit message. Please review.

shiroyasha:
In the meantime, I have found [this one comment](https://bugs.ruby-lang.org/issues/11788#note-1) that suggest that the combination of `iseq.to_a + marshal` could be used. This would allow support for Ruby2.2 but unfortunately it comes with a performance penalty as described [here](https://bugs.ruby-lang.org/issues/11788#note-2).

burke:
Thanks!

jules2689:
LGTM! Thanks!

As for https://github.com/Shopify/bootsnap/pull/18#issuecomment-298023585, I think that perf hit would be pretty brutal. I'd be ok with evaluating a POC, but the complexity may not be the best to add. Thank you for looking into this!

PR #21, opened 2017-04-30
FEATURE: stop using xattr binary to set/remove xattrs

SamSaffron:
Using ffi-xattr has better compatibility with Linux

### DIFF
--- a/bootsnap.gemspec
+++ b/bootsnap.gemspec
@@ -30,6 +30,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake-compiler', '~> 0'
   spec.add_development_dependency "minitest", "~> 5.0"
   spec.add_development_dependency "mocha", "~> 1.2"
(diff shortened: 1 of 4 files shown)


shiroyasha:
Could we also use `ffi-xattr` to to write the implementation in Ruby? :thinking: 

Or is there a known performance penalty if we do something like this:

``` ruby
binary = RubyVM::InstructionSequence.compile_file(path).to_binary

xattr = Xattr.new(path)
xattr["user.aotcc.value"] = binary
```

SamSaffron:
Yeah, there will be lots of extra RVALUES generated cause the actual lookup first looks up metadata that would end up being about 6-7 Ruby objects and then there is crc and all sort of other jazz.

burke:
I tried a bunch of less-aggressively-C implementations before settling on the one we have now. Performance sucked for the reasons that Sam outlined, and also because it's difficult to convince ruby that it doesn't have to do a huge number of additional syscalls.

Issue #14, opened 2017-04-25
compile_cache_iseq and compile_cache_yaml on Linux


SamSaffron:
I wanted to open a placeholder topic for getting support for `compile_cache_iseq` and `compile_cache_yaml` on Linux. 

In general (when I am not travelling) Linux is my dev box so I would love to have everything working there. 

What are the blockers here? What needs to be implemented?

burke:
The code right now uses extended filesystem attributes on the source files as the cache backend. The code actually supports linux in a "theoretically it might work" kind of way, but linux doesn't mount volumes with `-o user_xattr` by default. It's totally possible it would work with 0-5 lines changed on a volume with `user_xattr`.

Probably the compile_cache stuff should be rewritten to use some other backend. I couldn't find anything as fast as this, but I'd trade a few ms for broader compatibility.

SamSaffron:
Hmm, brainstorming, what about an optional redis or memcached store?

burke:
I'm pretty sure the network overhead would kill it. I tried it with [lmdb](https://github.com/minad/lmdb) which basically `mmap`s a b-tree in and returns values as strings without copying them out of the mmap'd region. That was a tiny bit slower than this on read, and quite a lot slower on insert. Redis and memcached would be at best comparable to this, but with additional network overhead.

It probably makes sense to just have an alternate strategy that uses a cache directory, like `tmp/cache/bootsnap-compile-cache/de/deadbeefcafe00042` or whatever, maybe key it on md5sum of the filename or something. This would be more container-friendly too.

wjordan:
> I tried it with lmdb which basically mmaps a b-tree in and returns values as strings without copying them out of the mmap'd region. That was somewhat slower than this on read, and quite a lot slower on insert.

The [lmdb](https://github.com/minad/lmdb) gem seems to have _terrible_ write performance, according to [moneta](https://github.com/minad/moneta)'s CI benchmarks (and my own ad-hoc tests)- one [example benchmark run](https://travis-ci.org/minad/moneta/jobs/228354777) shows only **~700** write ops/sec compared to **~160k** ops/sec for [LocalMemCache](https://github.com/sck/localmemcache) in the same test, or **> 200x** faster).

> Redis and memcached would be at best comparable to this, but with additional network overhead.

[...]

PR #25, opened 2017-05-03
extract cache implementation from library

wjordan:
This PR simplifies the caching implementation by extracting the (OSX-only) C extension and third-party serialization/compression dependencies (snappy / msgpack), and instead accepting any cache object that implements one of several common interfaces (`set`/`get`, `read`/`write`, etc).

I've changed the public API slightly so the `load_path_cache`, `compile_cache_iseq`, and `compile_cache_yaml` config options each accept an existing instantiated `cache` object (`ActiveSupport::Cache::FileStore.new('tmp/cache')` as an example, if using Active Support) instead of a boolean value. The optimizations can all use the same cache or separate ones depending on the particular use-case.

The cache interface is also compatible with [`moneta`](https://github.com/minad/moneta) cache objects, making this library compatible with a large number of cache implementations that could be used along with this PR. They can be swapped around and compared against one another in macro benchmarks to determine optimal performance for a given application.

At this point this is just a proof-of-concept for discussion and comparison. I'm curious to see some macro-benchmarks comparing boot times using a typical `FileStore` cache against the original xattr-based implementation.


burke:
I'm ok with this at a conceptual level but I don't think I'll have time to read through the code until next week.

jules2689:
I like the concept, but I think I'd like to explore if it's possible for the bootsnap cache to remain as an option without an additional library.

lib/bootsnap/cache_wrapper.rb
```diff
@@ -0,0 +1,55 @@
+# Wrapper to accept a number of common cache implementations.
+module Bootsnap
+  class CacheWrapper
+    def self.get(cache)
+      # `Cache#get(key)` for Memcache
+      if cache.respond_to?(:get)
```
SamSaffron: I really dislike this pattern, the cache bootsnap takes in should always require a consistent interface not a "anything goes" one, otherwise it is #respond_to central for zero gain.

SamSaffron:
something about the implementation here is not right 

removing msgpack is premature here, the current pattern of a single file for the entire load path is way better than requiring 255 files for the load path cache

In fact this branch is slower overall than current for Discourse even with iseq cache, I measured the iseq cache to save approx 100ms of bootime for Discourse.

But, this branch with everything enabled:

```
bin/rails r 'puts'  2.37s user 0.26s system 98% cpu 2.679 total
sam@ubuntu discourse % time bin/rails r 'puts'

bin/rails r 'puts'  2.38s user 0.23s system 98% cpu 2.660 total
sam@ubuntu discourse % time bin/rails r 'puts'

bin/rails r 'puts'  2.41s user 0.20s system 98% cpu 2.664 total
sam@ubuntu discourse % time bin/rails r 'puts'

bin/rails r 'puts'  2.42s user 0.21s system 98% cpu 2.683 total
```

This branch with load path only:

```
bin/rails r 'puts'  2.43s user 0.31s system 98% cpu 2.770 total
sam@ubuntu discourse % time bin/rails r 'puts'

bin/rails r 'puts'  2.48s user 0.30s system 98% cpu 2.810 total
sam@ubuntu discourse % time bin/rails r 'puts'

bin/rails r 'puts'  2.54s user 0.30s system 97% cpu 2.911 total
sam@ubuntu discourse % time bin/rails r 'puts'

bin/rails r 'puts'  2.49s user 0.28s system 98% cpu 2.809 total
```

Current on Linux

```
bin/rails r 'puts'  2.40s user 0.23s system 98% cpu 2.672 total
sam@ubuntu discourse % time bin/rails r 'puts'

bin/rails r 'puts'  2.44s user 0.19s system 98% cpu 2.668 total
sam@ubuntu discourse % time bin/rails r 'puts'

bin/rails r 'puts'  2.40s user 0.29s system 98% cpu 2.721 total

```

I think we need to be a lot more measured here, leave the current implementation alone and introduce a switch to enable an alternative cache for iseq/yaml on linux, leave all the current c extension alone as well.

burke:
The 0.3.1 cache implementation isn't darwin-specific, so there shouldn't be a need for this anymore.

wjordan:
@burke Excellent work in #43 replacing the xattr to file-based storage. That resolves the most urgent issue addressed by this PR, which was linux support (#14).

However, extracting the cache-implementation part from the boothook-integration part of this library still addresses a few other (definitely lower-priority) issues that I'd like to continue to raise for discussion:

- Support Ruby-only cache implementations without hard dependencies on C-extensions (e.g., for JRuby/Truffleruby support)
- Making the optimized, file-content-based cache implementation available in other contexts through a generic `#fetch(path, &block)` interface, e.g.:

```ruby
processed_file = file_cache.fetch(path) do |file_contents|
  expensive_process(file_contents)
end
```
- Support customizing the combination of serializer (`msgpack`) / compressor(`snappy`) / persistence-layer (filesystem) used by the boothook cache (to allow experimentation, and to customize/optimize for specific use-cases)
- Reduce unnecessary complexity (and development/maintenance burden) of the library. (That is, all other things being equal, extracting the highly-optimized C-extension cache-implementation into a separate project would make this boothook project much less complex, much more stable/robust and much easier to develop/maintain moving forward.)

I'll reopen another PR based off the latest 0.3.x code that will continue to address these remaining issues (and offer some benchmark numbers for performance comparison).

jules2689:
Fantastic! Thanks for the excellent post @wjordan

SamSaffron:
My #1 concern here is that I don't want us to pay any performance price for flexibility. The current implementation is very well tuned.

jules2689:
@SamSaffron agreed, but @wjordan also mentioned benchmarks, so let's wait for those :)

burke:
Right now, we call `Bootsnap::CompileCache::Native.fetch(dir, path, handler)`. If the backend was swappable but supported that same API, I don't think there would be any performance impact, and I wouldn't have much issue with it as long as the complexity increase wasn't too high.

wjordan:
Please take a look at #56 for the followup on this. [Benchmarks](https://gist.github.com/wjordan/91f577f5af06c5e05a9c083dc3ccd5a3) from my local machine show similar performance (and ~70 disk-space usage improvement via `snappy`) loading the `discourse` app.

PR #26, opened 2017-05-03
ignore relative paths added to $LOAD_PATH

wjordan:
This PR rescues the `RelativePathNotSupported` error when a relative path is added to the `$LOAD_PATH`.

The use case is an application including the `cucumber` gem, which has [code](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/load_path.rb#L14) that unshifts a relative path `lib` onto `$LOAD_PATH`.

The desired behavior would be to just skip caching files loaded via a relative load path, rather than raising an error that exits the application. There are likely some side-effects from this change (I don't understand the load-path caching code very well) so please let me know if this workaround is misguided for some reason.

### DIFF
--- a/lib/bootsnap/load_path_cache/cache.rb
+++ b/lib/bootsnap/load_path_cache/cache.rb
@@ -121,7 +121,11 @@ module Bootsnap
           paths.map(&:to_s).each do |path|
             p = Path.new(path)
             next if p.non_directory?
-            entries, dirs = p.entries_and_dirs(@store)
+            begin
+              entries, dirs = p.entries_and_dirs(@store)
+            rescue PathScanner::RelativePathNotSupported
+              next
+            end
             # push -> low precedence -> set only if unset
             dirs.each    { |dir| @dirs[dir]  ||= true }
             entries.each { |rel| @index[rel] ||= path }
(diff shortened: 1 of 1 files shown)


burke:
Yeah, the problem is that bootsnap asserts that its `$LOAD_PATH` knowledge is complete, and returns negative results (i.e. raises `LoadError`s) when asked for a file that it thinks is not on the path.

So we can't actually ignore paths.

What we could do is to resolve the path to an absolute path, but we'd have to either:

1. Deal with the fact that it would all be lies if the application changes its cwd.
2. Handle relative paths "properly", and propagate an event when the application changes directory, which causes a re-scan of any relative paths.

It's a kind of thorny issue, which is why it just raises an exception. In our applications, we either already didn't use relative paths, or they were exceedingly easy to change to absolute paths.

wjordan:
> Yeah, the problem is that bootsnap asserts that its $LOAD_PATH knowledge is complete, and returns negative results (i.e. raises LoadErrors) when asked for a file that it thinks is not on the path.

Is this behavior necessary for some reason? As an alternative, could bootsnap be made to _not_ assert that its `$LOAD_PATH` knowledge is complete, and instead delegate to the default implementation when asked for a file that it thinks is not on the path?

If the concern with this alternative is a possible performance impact in the case where `LoadError` is raised often, the fallback behavior could be made opt-in via config flag rather than default. I'd much rather opt-in to slightly reduced performance for `LoadError` case (especially if, for example, I know my application does not raise many `LoadErrors` internally on startup), versus the whole feature not working at all for applications containing relative paths.

burke:
https://github.com/Shopify/bootsnap/pull/49 addresses this.

PR #27, opened 2017-05-05
FEATURE: remove libsnappy as it is not saving much

SamSaffron:
Almost exactly the same load time for the caches with or without compression

By eliminating libsnappy we reduce dependencies and make it easier to adopt bootsnap

### DIFF
--- a/bootsnap.gemspec
+++ b/bootsnap.gemspec
@@ -33,5 +33,4 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "ffi-xattr", "~> 0.1.2"
 
   spec.add_runtime_dependency "msgpack", "~> 1.0"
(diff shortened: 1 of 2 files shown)


burke:
This works for me. I guess this has a nice self-healing property too since it handles the malformed messagepack error by just regenerating the cache, yeah?

SamSaffron:
yeah, when I tested the change the message pack error popped up, nice to self-heal from corrupt cache.

SamSaffron:
@burke do you mind doing another release? I would like to add bootsnap to discourse but libsnappy was kind of blocking us cause some devs had to do extra work to install Discourse on latest Ubuntus

burke:
Sorry for the delay; I took a few days off. I've pushed 0.2.13 now.

SamSaffron:
No problem at all, thanks!

PR #31, opened 2017-05-11
FIX: silently recover from cache corruption

SamSaffron:
cache could be corrupted due to libsnappy removal, this will silently recover from another case of malformed message pack data.


burke:
pushed 0.2.14

PR #37, opened 2017-05-19
fix if ActiveSupport is not defined

TheRusskiy:
otherwise may have issues in rake tasks
related to https://github.com/Shopify/bootsnap/pull/36


TheRusskiy:
@burke please merge this

PR #39, opened 2017-05-22
fix test_no_write_permission

wjordan:
Fixes `test_no_write_permission` unit test, by setting the `OPT_AOT_LOG` environment variable in the test (see passing [Travis CI build](https://travis-ci.org/wjordan/bootsnap/builds/234666264)).

Currently depends on Travis CI config (see separate PR #38), though that commit can be removed if desired.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+os: osx
+language: ruby
+rvm: ruby-2.4.0
+before_script: rake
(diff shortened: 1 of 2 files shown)


burke:
Ah, this has been failing for a long while and I'd just never bothered to look into it. Thanks very much for fixing it :)

PR #45, opened 2017-05-26
Add a message to relative path error

lime:
Setting up bootsnap on a project that uses Cucumber, I ran across the same issue described in https://github.com/Shopify/bootsnap/pull/26. Cucumber adds a relative `lib` load path, which bootsnap does not support.

I thought it would make debugging easier if the error message for `RelativePathNotSupported` included the offending path.

As a user, you are most likely to encounter this error when setting up bootsnap for the first time on an existing project. This message immediately tells you what to start grepping for to resolve the issue. :)


burke:
https://github.com/Shopify/bootsnap/pull/49 implements support for relative paths.

PR #56, opened 2017-06-05
Abstract cache interface

wjordan:
This is a followup to #25.

### Summary

This PR replaces the C-extension compile-cache implementation with an abstract cache interface that allows any cache implementation to be (optionally) provided to `Bootsnap#setup`.

By default, it uses a (slightly modified) existing `load_path_cache/store` implementation, which persists a serialized/compressed Ruby hash to a single file when the process exits.

### Benchmarks

I ran some measurements of `discourse` load-times on my local Ubuntu system comparing this PR against the current `master` branch, and there was little performance difference, while the use of `snappy` to persist the cache-storage file results in substantial disk-space savings:

Average of 5 test-runs:
`extract_cache` branch: **2.893ms** save; **1.708ms** load; **13M** disk space
`master` branch: **2.957ms** save; **1.716ms** load; **45M** disk space

(Detailed output in [gist](https://gist.github.com/wjordan/91f577f5af06c5e05a9c083dc3ccd5a3).)


wjordan:
I started some work on supporting the native c-extension fetch-cache (as opposed to removing it entirely) in another branch: [`extract_cache_c_ext`](https://github.com/wjordan/bootsnap/compare/extract_cache...wjordan:extract_cache_c_ext?expand=1) - I had difficulty getting it to work reliably without segfaults, possibly related to the slightly modified interface - the iseq compile cache was segfaulting on very-large sources ([`certificate_generator.rb`](https://github.com/discourse/discourse/blob/master/plugins/discourse-narrative-bot/lib/discourse_narrative_bot/certificate_generator.rb) in `discourse`), and it was having all sorts of 'invalid binary' issues unless you run msgpack before storing the iseq binary.

wjordan:
@burke please take a look and let me know if you're willing to even consider merging a PR along these lines. Happy to engage in further discussion or follow-up work if you have specific code comments or concerns with the general approach.

burke:
Just getting back from a vacation. I'll take a look at this in the next little bit, lots to catch up on. I'm not immediately opposed to the idea.

wjordan:
Closing due to lack of further interest in this contribution.

Issue #52, opened 2017-06-01
Windows support


daniel-rikowski:
In the light of #50 I wonder if the missing support for Windows is a final decision or if you would **accept pull requests for Windows support?** (At least partial support, see below)  

I'm asking because I could get the path pre-scanning feature working on Windows with [Ruby 2.4.1 (using MSYS2)](https://github.com/oneclick/rubyinstaller2) in about 30 minutes and some minor code changes. (Mostly making the C code compile and fixing two root path detection lines in Ruby)

The speed improvements were dreamlike!  On a fast Windows workstation I could bring the startup time of my Rails app down to 10 seconds from 30 seconds and on a very slow laptop down to 25 seconds from originally 2 minutes. (🐌 -> 🚀)

The compilation caching did create the cache files, but reading them results in a "bad file descriptor" error in the C code, which I couldn't track down (yet). But still, the speed improvements of the path scanning alone warrant the use of bootsnap on Windows platforms. (IMO)

The only downside is detecting an absolute path: Under a Unix-like platform it's just checking for a leading slash, but on Windows it's a little more involved. Currently I'm checking `path[1..2] == ':/'` but that might not be sufficient and - perhaps more importantly - slows down the check on all platforms. 

In any case: Thank you for this awesome gem!

joostliketoast:
not directly relevant to this gem. But a good solution could be to run WSL which is all sorts of amazing.
it runs my rails productions environments without a hitch( File I/O is a bit slow at the moment but the next release should solve that)

jules2689:
We would entertain the possibility of windows support :) PRs are welcome and we can discuss the best course of action.

It is absolutely crucial that it doesn't impact unix platforms like you mention - but we could perhaps use a separate C file for windows? I'm not sure how gems normally handle windows.

daniel-rikowski:
The recent changes to bootsnap made it a lot easier to maintain Windows compatibility. See #60 
With this I could use all bootsnap features, even the compilation caching.

Here's what I did:

On the Ruby side only one place needed a change to the absolute path detection. I implemented it by checking for a colon at the second character. This is not perfect, but someone using UNC paths most likely isn't concerned with speed. Unfortunately I had to add a separate method call, but the performance hit should be negligible. (🤞)

On the C side I had to mark all open file descriptors as binary and replace a mkdir call which has no mode parameter on Windows. I also return 0 for the OS version hash. I'm not sure *what* to return here, I guess it could be some internal MSYS version marker.

[...]

PR #60, opened 2017-06-17
Added preliminary Windows compatibility

daniel-rikowski:
These changes make it possible to use bootsnap on Windows. See #52 for discussion.

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -179,7 +179,7 @@ static uint32_t
 get_os_version(void)
 {
   #ifdef _WIN32
-  /* Ignore for now */
+  return (uint32_t)GetVersion();
   #else
   uint64_t hash;
   struct utsname utsname;
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -18,7 +18,9 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <sys/stat.h>
+#ifndef _WIN32
 #include <sys/utsname.h>
+#endif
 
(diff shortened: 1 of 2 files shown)


burke:
This is much less extra code than I expected. Awesome! Definitely ok with merging this if you can figure out some reasonably sensible implementation for the `os_version` field.

daniel-rikowski:
At the very least it should return zero. That's an obvious bug... 😨 

I'm now returning the result of `GetVersion`, it contains the major and minor version and the build number of Windows, squashed into a uint32_t.

jules2689:
Yea, thanks a lot for this! Glad this can be expanded to all windows devs too!

janpio:
Can/should Windows support be added to the README?

Inversion-des:
How should we use it on Windows?
`supported?` returns `false` due to `RUBY_PLATFORM =~ /darwin|linux|bsd/`

I tested that for my app on Ruby 2.6.3 `compile_cache_iseq: true` works and reduces loading time for -36%, which is great.

Tried also `load_path_cache: true` but it just stuck with 100% CPU usage.

daniel-rikowski:
The `supported?` check disabling Bootsnap on Windows was added about 18 months after the actual Windows support was merged. I'm not sure why that was added, I assume the intention was to "formalize" the default setup, which never enabled the compile cache on Windows. (Although no reason for that after Windows was added) (https://github.com/Shopify/bootsnap/commit/464104800277cb92022701608095583c3e1cc725)

I've been using Bootsnap on Windows ever since, now on Ruby 2.7.1 and Rails 6. (with `supported?` changed to return true on Windows)

@Inversion-des I don't know what is causing that, it should work. Whenever I suspect any Bootsnap "weirdness" I just delete the Bootsnap cache (on Rails it is in `tmp/cache`)

PR #61, opened 2017-06-17
streamline working and warn about iseq caching

grosser:
@burke

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,10 @@
 
 **Beta-quality. See [the last section of this README](#trustworthiness).**
 
-Bootsnap is a library that plugs into a number of Ruby and (optionally) `ActiveSupport` and `YAML`
-methods to optimize and cache expensive computations. See [the How Does This Work section](#how-does-this-work) for more information.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,10 +2,10 @@
 
 **Beta-quality. See [the last section of this README](#trustworthiness).**
 
-Bootsnap is a library that plugs into a number of Ruby and (optionally) `ActiveSupport` and `YAML`
-methods to optimize and cache expensive computations. See [the How Does This Work section](#how-does-this-work) for more information.
+Bootsnap is a library that can plug into Ruby, `ActiveSupport` and `YAML`
```
jules2689: We want to emphasize that ActiveSupport is optional. Rails is not a dependency for Bootsnap

grosser:
all of them are optional, so put them on equal footing ?

jules2689:
> Bootsnap is a library that plugs into Ruby, with optional support for `ActiveSupport` and `YAML`

How about something like this?

grosser:
sounds good, updated

PR #63, opened 2017-06-19
Detect coverage earlier

burke:
@grosser @jules2689 

This should bail out of the iseq caching early enough. Testing whether the constant is defined is *very* fast. This should add less than 2ms to shopify boot.

Where this falls down is if coverage is eager-loaded by something, but not actually used.

The only ruby-land API available to us to test whether coverage is actually running is `Coverage.peek_result`, which appears to be quite expensive if coverage is currently running. I think that's unviable. I wonder if we can add a little function to the C extension to check this more economically.

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -111,6 +111,15 @@ struct s2o_data;
 struct i2o_data;
 struct i2s_data;
 
+/* https://bugs.ruby-lang.org/issues/13667 */
+extern VALUE rb_get_coverages(void);
+static VALUE
+bs_rb_coverage_running(VALUE self)
+{
+  VALUE cov = rb_get_coverages();
+  return RTEST(cov) ? Qtrue : Qfalse;
+}
+
 /*
  * Ruby C extensions are initialized by calling Init_<extname>.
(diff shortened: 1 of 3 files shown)


burke:
`rb_get_coverages` exists, though not in `ruby.h`. I `extern`'d it and this seems to work fine. I'll make a ruby issue too, it would be nice to have `Coverage.running?`

grosser:
oh, sneaky :)

burke:
https://bugs.ruby-lang.org/issues/13667

jules2689:
So this would add 1ms of time to check for coverage? 💩  I guess it's needed though

burke:
```
> Benchmark.realtime { $LOADED_FEATURES.size.times { defined?(Coverage) && Bootsnap::CompileCache::Native.coverage_running? } } - Benchmark.realtime { $LOADED_FEATURES.size.times { } }
=> 0.0005639999872073531
```

So yeah I'm not too worried about that.

jules2689:
Ah, ok. :+1:. 
I had just divided the 1.36s by 100 from [here](https://bugs.ruby-lang.org/issues/13667).

PR #64, opened 2017-06-19
Bundler.bundle_path considered STABLE

burke:
fyi @jules2689 @grosser


jules2689:
How long does `Bundler.bundle_path` take to run?

burke:
It's memoized: https://github.com/bundler/bundler/blob/d4a06723f2552f55bc46b395ab1cc0070b438b7e/lib/bundler.rb#L78-L80

jules2689:
Ah ok :+1:

PR #78, opened 2017-08-02
Update gem description and summary

frodsan:


burke:
Hah. Forgot I hadn't got around to that. Thank you!

PR #81, opened 2017-08-04
Fix warnings

tenderlove:
Hi,

This PR fixes some runtime / load time warnings and enables warnings on test.  I need this because enabling debug mode (running Ruby with `-d`) enables warnings, and bootsnap is very noisy with warnings enabled.

Thanks!


rafaelfranca:
❤️

PR #82, opened 2017-08-25
Use reverse_each

tcopeland:


tommotorefi:
Cool yeah and the beauty of it is that https://github.com/tcopeland/pippi#reversefollowedbyeach found it!  Always nice when the tools find something...

tcopeland:
@burke also let me just seize this opportunity to say thanks very much for bootsnap.  It has saved me so much time and annoyance; seeing a `rails c` open up super-quick makes my day that much brighter.  Thank you!

burke:
Thanks, glad to hear it 😄

lib/bootsnap/load_path_cache/cache.rb
```diff
@@ -152,7 +152,7 @@ def push_paths_locked(*paths)
 
       def unshift_paths_locked(*paths)
         @store.transaction do
-          paths.map(&:to_s).reverse.each do |path|
+          paths.map(&:to_s).reverse_each do |path|
             p = Path.new(path)
```
scottbartell: This is totally unrelated but I think it might not be a good idea to use `p` as a variable name. `p` is a method defined in the Kernel module https://ruby-doc.org/core-2.4.1/Kernel.html#method-i-p

burke: I don't have much of a problem with it; it's clear from context that it's not being used as a function, nor would we be pretty-printing things to stdout from here.

burke:
Apparently I approved this but forgot to merge it. Whoops!

PR #83, opened 2017-08-28
Follow symlinks while traversing LOAD_PATH

barthez:
When I was trying to setup bootsnap in our project I had troubles with loading `psych/exception` and `bigdecimal/util` files. It turned out we have `psych` and `bigdecimal` directories *symlinked* in `/usr/share/ruby` dir and bootsnap does not follow symlinks. It is how ruby gems are provided on CentOS.

### DIFF
--- a/lib/bootsnap/load_path_cache/path_scanner.rb
+++ b/lib/bootsnap/load_path_cache/path_scanner.rb
@@ -1,8 +1,18 @@
+require_relative '../explicit_require'
+
 module Bootsnap
   module LoadPathCache
     module PathScanner
-      REQUIRABLES_AND_DIRS = "/**/*{#{DOT_RB},#{DL_EXTENSIONS.join(',')},/}"
-      IS_DIR = %r{(.*)/\z}
+      # Glob pattern to find requirable files and subdirectories in given path.
+      # It expands to:
+      #
+      #   * `/*{.rb,.so,/}` - It matches requirable files, directories and
+      #     symlinks to directories at given path.
+      #   * `/*/**/*{.rb,.so,/}` - It matches requirable files and
+      #     subdirectories in any (even symlinked) directory at given path at
+      #     any directory tree depth.
+      #
+      REQUIRABLES_AND_DIRS = "/{,*/**/}*{#{DOT_RB},#{DL_EXTENSIONS.join(',')},/}"
       NORMALIZE_NATIVE_EXTENSIONS = !DL_EXTENSIONS.include?(LoadPathCache::DOT_SO)
       ALTERNATIVE_NATIVE_EXTENSIONS_PATTERN = /\.(o|bundle|dylib)\z/
       BUNDLE_PATH = (Bundler.bundle_path.cleanpath.to_s << LoadPathCache::SLASH).freeze
@@ -27,12 +37,13 @@ module Bootsnap
           next if contains_bundle_path && absolute_path.start_with?(BUNDLE_PATH)
           relative_path = absolute_path.slice!(relative_slice)
 
-          if md = relative_path.match(IS_DIR)
-            dirs << md[1]
+          if relative_path.end_with?('/')
+            dirs << relative_path[0..-2]
           else
             requirables << relative_path
           end
         end
+
         [requirables, dirs]
       end
     end
--- a/test/load_path_cache/path_scanner_test.rb
+++ b/test/load_path_cache/path_scanner_test.rb
@@ -1,3 +1,5 @@
+require 'test_helper'
+
 module Bootsnap
   module LoadPathCache
     class PathScannerTest < MiniTest::Test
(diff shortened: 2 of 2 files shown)


barthez:
I've added a bit of performance improvement since operating on arrays if we need a structure that's containing unique elements can be inefficient. Also, differing dirs from files in glob output can be done with `end_with?` instead of regexp matching, which is slower.

lib/bootsnap/load_path_cache/path_scanner.rb
```diff
@@ -20,20 +23,23 @@ def self.call(path)
         # and the bundle path is '.bundle'.
         contains_bundle_path = BUNDLE_PATH.start_with?(path)
 
-        dirs = []
-        requirables = []
+        dirs = Set.new
```
burke: I have mixed feelings on this. On one hand, 

> [`Set#to_a` Converts the set to an array. The order of elements is uncertain.](https://ruby-doc.org/stdlib-2.4.1/libdoc/set/rdoc/Set.html#method-i-to_a)

...but on the other hand, 

> [Hashes enumerate their values in the order that the corresponding keys were inserted.](https://ruby-doc.org/core-2.4.1/Hash.html)

...and...

> [Set uses Hash as storage, ...](https://ruby-doc.org/stdlib-2.4.1/libdoc/set/rdoc/Set.html)

So while it's relatively clear that this is in fact safe, I don't like that it relies on behaviour that is explicitly undefined.

lib/bootsnap/load_path_cache/path_scanner.rb
```diff
@@ -1,8 +1,11 @@
+require_relative '../explicit_require'
+
+Bootsnap::ExplicitRequire.from_rubylibdir('set')
+
 module Bootsnap
   module LoadPathCache
     module PathScanner
-      REQUIRABLES_AND_DIRS = "/**/*{#{DOT_RB},#{DL_EXTENSIONS.join(',')},/}"
-      IS_DIR = %r{(.*)/\z}
+      REQUIRABLES_AND_DIRS = "/**{,*/**}/*{#{DOT_RB},#{DL_EXTENSIONS.join(',')},/}"
```
burke: So expanding the first curlies, you get:

```ruby
"/**/*{#{DOT_RB},#{DL_EXTENSIONS.join(',')},/}"
"/***/**/*{#{DOT_RB},#{DL_EXTENSIONS.join(',')},/}"
```

Am I missing something? I've never seen a triple asterisk before and I can't find any references for it. I don't understand why this change would make it resolve symlinks. I'm probably just missing some weird bit of arcana on how glob works?

At any rate, this pattern was already verging on too complex for one line and no documentation, I think it will probably need to be written more clearly somehow if we're going to add more complexity to it like this.

lib/bootsnap/load_path_cache/path_scanner.rb
barthez: Order of either dirs or requirables is irrelevant. The only place where it is used is [Cache](https://github.com/Shopify/bootsnap/blob/master/lib/bootsnap/load_path_cache/cache.rb#L147-L148) where they are transformed to hash. The uncertainty of the order of `Set#to_a` is even more irrelevant if you know that order of entries yielded from `glob` is uncertain as well.

Let me paste benchmark comparing current implementation with the previous one:
```
require 'bootsnap'
require 'benchmark/ips'

Benchmark.ips do |b|
  b.report('after') { Bootsnap::LoadPathCache::PathScanner.call($LOAD_PATH.last) }
  b.report('before') { Bootsnap::LoadPathCache::PathScanner.call($LOAD_PATH.last) }

  b.hold! __FILE__.sub('.rb', '.json')
  b.compare!
end

# Warming up --------------------------------------
#               before    15.000  i/100ms
# Calculating -------------------------------------
#               before    153.094  (± 3.9%) i/s -    765.000  in   5.005778s
# 
# Comparison:
#                after:      484.5 i/s
#               before:      153.1 i/s - 3.16x  slower
```

lib/bootsnap/load_path_cache/path_scanner.rb
barthez: That's my bad. I have missed one `/` and it for some reason worked ¯\_(ツ)_/¯.  I have found better glob pattern to find everything we need without duplicates (I can get back to arrays instead of sets).

lib/bootsnap/load_path_cache/path_scanner.rb
barthez: Scratch that, I have found better glob pattern that does not yield duplicates, sets are not necessary anymore.

barthez:
@burke Thanks for your review!

I've changed glob pattern. I hope is it more clear now, I've added some docs as well. With this glob pattern, we don't need sets anymore since `Dir.glob` does not yield duplicates.

burke:
Ah, ok, this is much more clear. I'll try this out on one of our large projects today.

barthez:
Any news on that @burke?

burke:
Got distracted last week, I'll get back to it this week.

burke:
Ok, looks good! Sorry for the delay, and thanks for the fix!

burke:
(released 1.1.3 with this change)

PR #84, opened 2017-08-28
Fix typo in README

Neodelf:


burke:
Huh, wonder why I did that. Thanks :+1:

PR #85, opened 2017-09-10
Allow preprocessing content before it is cached

kddeisz:
Since bootsnap is already hooking into the compile chain, this additionally allows the files to be preprocessed before they are compiled. This opens up a bunch of different possibilities, like macros or AST rewriting. Mostly this allows me to hook it into my own preprocessing gem (http://github.com/kddeisz/vernacular). Let me know what you think and what else might need to happen to get this merged.

### DIFF
--- a/lib/bootsnap.rb
+++ b/lib/bootsnap.rb
@@ -12,7 +12,8 @@ module Bootsnap
     autoload_paths_cache: true,
     disable_trace: false,
     compile_cache_iseq: true,
-    compile_cache_yaml: true
+    compile_cache_yaml: true,
+    preprocesser: nil
   )
     if autoload_paths_cache && !load_path_cache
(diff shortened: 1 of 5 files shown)


burke:
This is a really interesting use-case, but I'm not super inclined to merge this change for a combination of two reasons:

1. This would be useful to a very small fraction of bootsnap's users;
2. It should be trivially easy to accomplish the same thing by monkeypatching `CompileCache::{ISeq,YAML}.input_to_storage`.

I'd encourage you to just monkeypatch it and suggest any downstream users do the same. The code it would depend on is very unlikely to change in the future. If the bootsnap+vernacular combination gains a large number of users via monkeypatch, I'd consider merging something like this in the future (i.e. if condition 1 was no longer true).

kddeisz:
Makes total sense. Thanks for the thoughtful reply!

PR #88, opened 2017-10-05
Avoid loading a constant twice by checking if it is already defined

### LINKED ISSUE #74: `with_bootsnap_fallback`: Constants are loaded twice when a `NameError` occurs
### Problem

`NameError` during `load_missing_constant` results in constants being loaded twice.

### Investigation

Take the following basic example, which can be found in [this repo](https://github.com/xuorig/bootsnap-issue/blob/master/example.rb) also:

```ruby
# example.rb
a = Bar.new
```

```ruby
# Bar.rb
class Bar
  DoesNotExist
end
```
 
  The following methods are called when `Bar` is first encountered:

  - `ActiveSupport::Dependencies#const_missing `
  - `Bootsnap#load_missing_constant`
  - `Bootsnap#with_bootsnap_fallback`
  - `super` => `ActiveSupport::load_missing_constant from_mod: Object, const_name: Bar`
  -  ...

Once Bar is loaded, `DoesNotExist` now has to be loaded, the same methods are called for `DoesNotExist` but remember this is all wrapped in the original `Bootsnap#load_missing_constant` and `Bootsnap#with_bootsnap_fallback`.

Since `DoesNotExist` cannot be found, a `NameError` is raised. It is reraised through the stack until it hits the rescue in [with_bootsnap_fallback](https://github.com/Shopify/bootsnap/blob/master/lib/bootsnap/load_path_cache/core_ext/active_support.rb#L7-L12).

The exception is caught, and then `yield` is called again because the loading is tried this time `without_bootsnap_cache`. This means the orignial `load_missing_constant` for `DoesNotExist` is called again. This of cource raises a `NameError` again, which this time is not caught by this rescue. Instead, it is caught by an other `load_missing_constant` call, the original call to load the `Bar` class.

The same process is repeated, `yield` is called again, which causes `Bar` to be reloaded, executing any logic twice.

### Real life example

This causes issues when a loaded constant sets state on load. Take this example:

```ruby
class Foo
  class_attribute :items
  self.items = {}
  
  def self.add_item(name, klass)
   if items.key?(name)
     raise Error, "cannot add same item twice!!!"
   end
   
    self.items = items.merge(name: klass)
  end
  
  # This one will run ok
  add_item :a, Bar
  # This one raises and causes the double load
  add_item :b, DoesNotExist
end
```

When loading `DoesNotExist`, `DoesNotExist` is being loaded twice, which fails both times but `Foo` is also loaded twice since it rescues the `NameError`.

This causes `add_item, :a, Bar` to be called twice, which causes this error to be raised:

`raise Error, "cannot add same item twice!!!"`

In reality, we would've liked to see the `NameError` raised, which causes a really weird developer experience.

### TL;DR

In `load_missing_constant`, `NameErrors` are caught by Bootsnap's `with_bootsnap_fallback`, and retried without bootsnap cache. This causes ActiveSupport's `load_missing_constant` to be called twice on all constants that were loading / led to the `NameError`.

### Questions

  - Is this the expected behaviour?
  - My ruby internals knowledge is limited on this one, should constants be loaded twice like this, could we unload them before trying without bootsnap cache?
  - Is this something developers need to be mindful of? Seems like this could happen to anybody using the gem.

cc @jules2689, @burke, @rafaelfranca

dylanahsmith:
Fixes #74
cc @patrickdonovan since this just caused you some confusion today and would have given you the relevant error

`load_missing_constant` has a fallback for when the constant can't be loaded from the cache, so that it will try without using the cache.  However, it does this when it receives a `NameError` exception, which could happen from an unrelated bug that happen in the file that loads that constant.

The fix was to use another condition that checks if the constant is already defined, rather than assuming the NameError was from the `return from_mod.const_get(const_name)` line in ActiveSupport::Dependencies.load_missing_constant.

I've tested with @patrickdonovan's branch in Shopify and using the code in https://github.com/xuorig/bootsnap-issue to make sure this is fixed.  The test suite doesn't really cover the active support extensions right now.

### DIFF
--- a/lib/bootsnap/load_path_cache/core_ext/active_support.rb
+++ b/lib/bootsnap/load_path_cache/core_ext/active_support.rb
@@ -2,15 +2,6 @@ module Bootsnap
   module LoadPathCache
     module CoreExt
       module ActiveSupport
-        def self.with_bootsnap_fallback(error)
-          yield
-        rescue error => e
-          # NoMethodError is a NameError, but we only want to handle actual
-          # NameError instances.
-          raise unless e.class == error
-          without_bootsnap_cache { yield }
-        end
-
         def self.without_bootsnap_cache
           prev = Thread.current[:without_bootsnap_cache] || false
           Thread.current[:without_bootsnap_cache] = true
@@ -50,13 +41,21 @@ module Bootsnap
           # behaviour.  The gymnastics here are a bit awkward, but it prevents
           # 200+ lines of monkeypatches.
           def load_missing_constant(from_mod, const_name)
-            CoreExt::ActiveSupport.with_bootsnap_fallback(NameError) { super }
+            super
+          rescue NameError => e
+            # NoMethodError is a NameError, but we only want to handle actual
+            # NameError instances.
+            raise unless e.class == NameError
+            raise if from_mod.const_defined?(const_name)
+            without_bootsnap_cache { super }
           end
 
           # Signature has changed a few times over the years; easiest to not
           # reiterate it with version polymorphism here...
           def depend_on(*)
-            CoreExt::ActiveSupport.with_bootsnap_fallback(LoadError) { super }
+            super
+          rescue LoadError
+            without_bootsnap_cache { super }
           end
         end
       end


dylanahsmith:
> I guess `depend_on` can never raise `NoMethodError`?

We were only rescuing LoadError for `depend_on`, so we weren't rescuing NoMethodError before for it.

dylanahsmith:
Should I make a bugfix release so we can take advantage of this in Shopify?

burke:
yeah, I'll do it though, the gem needs to be built with jruby too

burke:
Done; 1.1.4 is released

lastgabs:
This broke our app:

```
/Users/gabi/.gem/ruby/2.3.3/gems/bundler-1.15.0/lib/bundler/runtime.rb:85:in `rescue in block (2 levels) in require': There was an error while trying to load the gem 'jsonapi-resources'. (Bundler::GemRequireError)
Gem Load Error is: undefined method `without_bootsnap_cache' for ActiveSupport::Dependencies:Module
Backtrace for gem load error is:
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/active_support.rb:58:in `rescue in depend_on'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/active_support.rb:56:in `depend_on'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:251:in `require_dependency'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/abstract_controller/helpers.rb:150:in `block in modules_for_helpers'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/abstract_controller/helpers.rb:145:in `map!'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/abstract_controller/helpers.rb:145:in `modules_for_helpers'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/action_controller/metal/helpers.rb:93:in `modules_for_helpers'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/abstract_controller/helpers.rb:109:in `helper'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/abstract_controller/helpers.rb:187:in `default_helper_module!'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/abstract_controller/helpers.rb:36:in `block in inherited'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/abstract_controller/helpers.rb:36:in `class_eval'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/abstract_controller/helpers.rb:36:in `inherited'
/Users/gabi/.gem/ruby/2.3.3/gems/actionview-5.1.3/lib/action_view/layouts.rb:217:in `inherited'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/action_controller/metal/rendering.rb:23:in `inherited'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/action_controller/metal/parameter_encoding.rb:8:in `inherited'
/Users/gabi/.gem/ruby/2.3.3/gems/actionpack-5.1.3/lib/action_controller/metal/params_wrapper.rb:228:in `inherited'
/Users/gabi/.gem/ruby/2.3.3/gems/jsonapi-resources-0.9.0/lib/jsonapi/resource_controller.rb:2:in `<module:JSONAPI>'
/Users/gabi/.gem/ruby/2.3.3/gems/jsonapi-resources-0.9.0/lib/jsonapi/resource_controller.rb:1:in `<main>'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:292:in `block in require'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:258:in `load_dependency'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:292:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/jsonapi-resources-0.9.0/lib/jsonapi-resources.rb:7:in `<main>'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:50:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:50:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/bundler-1.15.0/lib/bundler/runtime.rb:82:in `block (2 levels) in require'
/Users/gabi/.gem/ruby/2.3.3/gems/bundler-1.15.0/lib/bundler/runtime.rb:77:in `each'
/Users/gabi/.gem/ruby/2.3.3/gems/bundler-1.15.0/lib/bundler/runtime.rb:77:in `block in require'
/Users/gabi/.gem/ruby/2.3.3/gems/bundler-1.15.0/lib/bundler/runtime.rb:66:in `each'
/Users/gabi/.gem/ruby/2.3.3/gems/bundler-1.15.0/lib/bundler/runtime.rb:66:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/bundler-1.15.0/lib/bundler.rb:108:in `require'
/Users/gabi/src/github.com/Shopify/u2/config/application.rb:19:in `<main>'
/Users/gabi/src/github.com/Shopify/u2/config/environment.rb:4:in `require_relative'
/Users/gabi/src/github.com/Shopify/u2/config/environment.rb:4:in `<main>'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:292:in `block in require'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:258:in `load_dependency'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:292:in `require'
/Users/gabi/src/github.com/Shopify/u2/test/test_helper.rb:4:in `<main>'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:292:in `block in require'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:258:in `load_dependency'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:292:in `require'
/Users/gabi/src/github.com/Shopify/u2/test/unit/helpers/teams_helper_test.rb:1:in `<main>'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:292:in `block in require'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:258:in `load_dependency'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:292:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/railties-5.1.3/lib/rails/test_unit/runner.rb:50:in `block in load_tests'
/Users/gabi/.gem/ruby/2.3.3/gems/railties-5.1.3/lib/rails/test_unit/runner.rb:50:in `each'
/Users/gabi/.gem/ruby/2.3.3/gems/railties-5.1.3/lib/rails/test_unit/runner.rb:50:in `load_tests'
/Users/gabi/.gem/ruby/2.3.3/gems/railties-5.1.3/lib/rails/test_unit/runner.rb:39:in `run'
/Users/gabi/.gem/ruby/2.3.3/gems/railties-5.1.3/lib/rails/commands/test/test_command.rb:38:in `perform'
/Users/gabi/.gem/ruby/2.3.3/gems/thor-0.20.0/lib/thor/command.rb:27:in `run'
/Users/gabi/.gem/ruby/2.3.3/gems/thor-0.20.0/lib/thor/invocation.rb:126:in `invoke_command'
/Users/gabi/.gem/ruby/2.3.3/gems/thor-0.20.0/lib/thor.rb:387:in `dispatch'
/Users/gabi/.gem/ruby/2.3.3/gems/railties-5.1.3/lib/rails/command/base.rb:63:in `perform'
/Users/gabi/.gem/ruby/2.3.3/gems/railties-5.1.3/lib/rails/command.rb:44:in `invoke'
/Users/gabi/.gem/ruby/2.3.3/gems/railties-5.1.3/lib/rails/commands.rb:16:in `<main>'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/bootsnap-1.1.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:292:in `block in require'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:258:in `load_dependency'
/Users/gabi/.gem/ruby/2.3.3/gems/activesupport-5.1.3/lib/active_support/dependencies.rb:292:in `require'
```

lib/bootsnap/load_path_cache/core_ext/active_support.rb
```diff
@@ -50,13 +41,21 @@ def remove_constant(const)
...
           # 200+ lines of monkeypatches.
           def load_missing_constant(from_mod, const_name)
-            CoreExt::ActiveSupport.with_bootsnap_fallback(NameError) { super }
+            super
+          rescue NameError => e
+            # NoMethodError is a NameError, but we only want to handle actual
+            # NameError instances.
+            raise unless e.class == NameError
+            raise if from_mod.const_defined?(const_name)
+            without_bootsnap_cache { super }
```
rafaelfranca: Ruby will not know from where this method come. We need to be explicit about who is the caller.

burke:
yanked; I'll fix and push 1.1.5 tomorrow.

lib/bootsnap/load_path_cache/core_ext/active_support.rb
dylanahsmith: I remember fixing that, although I was also developing against the bundle checkout of the gem.  I'm not sure how I missed that in this PR.  Obviously the fix is to prefix it with `CoreExt::ActiveSupport.` like the other calls to it in this file.

PR #96, opened 2017-11-28
Simplify the code to compare versions

gmalette:
This is just to simplify the ruby version comparison check, because it took me 45 minutes to understand it. I'm not smart, I'm sorry. 

```
> Gem::Version.new("1.9.3") >= Gem::Version.new("2.3.0")
=> false
> Gem::Version.new("2.3.0") >= Gem::Version.new("2.3.0")
=> true
> Gem::Version.new("2.4.2") >= Gem::Version.new("2.3.0")
=> true
```

This works as far back as I could test (Ruby 1.8.7 using the `ree-1.8.7.2012.02` Docker image)


gmalette:
Tests are passing locally, do you understand what would make them fail on CI?

burke:
Seems unrelated; I'll fix it on another PR.

PR #101, opened 2017-12-04
Fix miscellaneous tests

burke:


olleolleolle:
`time_file` was _way_ more stable.

PR #121, opened 2018-01-24
$LOAD_PATH optimization

maschwenk:
First off, this gem is dope 💯  🏀  🔥  . I think I've found a way to make it even faster 📊  💯 

`require bundler/setup` runs `Bundler.setup` with no group arguments by default. If you have a very deep `$LOAD_PATH`, you can avoid polluting it with with gems that are not relevant to the current environment.

I'm not sure this will apply to every project, but considering this gem is all about boot speed, it feels like mentioning this is a good idea in the README.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,11 @@ Add `bootsnap` to your `Gemfile`:
 gem 'bootsnap', require: false
 ```
 
-If you are using Rails, add this to `config/boot.rb` immediately after `require 'bundler/setup'`:
+If you are using Rails, add this to `config/boot.rb` (and remove `bundler/setup` if you have it):
 
 ```ruby
+require 'bundler'
+Bundler.setup(:default, ENV.fetch('RAILS_ENV', :development)) # Makes sure your $LOAD_PATH only contains the necessary gems
 require 'bootsnap/setup'
 ```
 
--- a/README.md
+++ b/README.md
@@ -18,11 +18,15 @@ Add `bootsnap` to your `Gemfile`:
 gem 'bootsnap', require: false
 ```
 
-If you are using Rails, add this to `config/boot.rb` (and remove `bundler/setup` if you have it):
-
+If you are using Rails, add this to `config/boot.rb` immediately after require `bundler/setup` or a `Bundler.setup` call:
+```ruby
+ require 'bundler/setup'
+ require 'bootsnap/setup'
+```
+Protip: Using Bundler.setup like in the example below can _further_ improve boot time by reducing the `$LOAD_PATH` to only include gems for the given environment
 ```ruby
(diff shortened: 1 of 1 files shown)


maschwenk:
I guess this is somewhat obviated by [Path Pre-Scanning](https://github.com/Shopify/bootsnap#path-pre-scanning)? It will speed up the very first scan but make no difference on subsequent lookups AFAIK?

README.md
```diff
@@ -18,9 +18,11 @@ Add `bootsnap` to your `Gemfile`:
 gem 'bootsnap', require: false
 ```
 
-If you are using Rails, add this to `config/boot.rb` immediately after `require 'bundler/setup'`:
+If you are using Rails, add this to `config/boot.rb` (and remove `bundler/setup` if you have it):
```
jules2689: I think this might be a bit confusing to people. Why should they remove bundler/setup?

Maybe this:

If you are using Rails, add this to `config/boot.rb` immediately after `require 'bundler/setup'` or a `Bundler.setup` call:

```ruby
 require 'bundler/setup'
 require 'bootsnap/setup'
 ```

```ruby
require 'bundler'
Bundler.setup(:default, ENV.fetch('RAILS_ENV', :development)) # Adds only the necessary gems to $LOAD_PATH
require 'bootsnap/setup'
 ```

Protip: Using `Bundler.setup` like in the example above can help boot time by only loading gems from your Gemfile.

**Comment by rafaelfranca** [comment]
We do that by default for the `Bundler.require` call, I see no reason why not do with the `Bundler.setup` too.

File: README.md (hunk 1 above)
Comment (maschwenk) [reply README.md]: Coming back at this a few months later so bare with me... 😄 

Unless I'm misunderstanding something, you want to remove `bundler/setup` because otherwise it'll run the bare `Bundler.setup` again (https://github.com/bundler/bundler/blob/master/lib/bundler/setup.rb) without the group setting. Their recommendation is outlined here: http://bundler.io/v1.16/guides/bundler_setup.html. So really, I'd amend my commit to be
```ruby
require 'bundler'
Bundler.setup(:default, :ci)
```

maschwenk: @jules2689 just a bump on this?

jules2689: Ah Sorry! This must have gotten lost while I was moving!
I'll look now

jules2689: Oh yea, sorry I wasn't clear.
I'd would like to mention why we want to remove bundler/setup

agis:
Is there any progress here? @maschwenk do you have any numbers that demonstrate the performance benefits of this PR?

burke:
Closing stale PR, IIRC the performance gain wasn't really huge.

PR #126, opened 2018-02-08
Explicitly mention what `disable_trace` does and the side effects

dannyfallon:
We used to have something in the README about `disable_trace` hinting
that it was incompatible with any tools that used `trace` functionality
and that for that reason Shopify didn't really use it that much.

It was super-surprising to me when I added Bootsnap to suddenly have
engineer debug workflows affected by it. A better description of the
side effect of the setting might have helped me narrow it down quicker
or prevented me shipping it in the first place with traces disabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Bootsnap.setup(
   development_mode:     env == 'development', # Current working environment, e.g. RACK_ENV, RAILS_ENV, etc
   load_path_cache:      true,                 # Optimize the LOAD_PATH with a cache
   autoload_paths_cache: true,                 # Optimize ActiveSupport autoloads with cache
-  disable_trace:        true,                 # (Alpha) Set `RubyVM::InstructionSequence.compile_option = { trace_instruction: false }`
+  disable_trace:        env == 'production',  # Set `RubyVM::InstructionSequence.compile_option = { trace_instruction: false }`. Disables all interactive debuggers and TracePoint features.
   compile_cache_iseq:   true,                 # Compile Ruby code into ISeq cache, breaks coverage reporting.
   compile_cache_yaml:   true                  # Compile YAML into a cache
 )


dannyfallon:
Some follow-up questions related to this feature:

1. Is there anything we can do to better explain why users would want to disable trace instructions?
  1. Is there a performance benefit to not having trace instructions? What is it?
  1. Would it make sense to disable tracing instructionsonly when compiling instruction sequences (i.e. when `cache_compile_iseq` is true)? It was weird to me that Bootsnap turned off trace instructions even when I opted to not cache my compiled iseq; This impacts code that is loaded even after application boot and Bootsnap's remit is faster application booting.
1. Is this `(Alpha)` anymore? 
1. Why is this enabled by default? I think it would be valuable to have `env == "production"` as the default to not surprise a bunch of people who want tracing when working locally

burke:
The difference is generally 1-2% performance improvement on ruby-instruction-evaluation-bounded (i.e. nearly anything not i/o bound) workloads. trace instruction generation disabling is sort of orthogonal to iseq caching; it simply determines, when a file is parsed and translated to bytecode, whether the generated code contains trace instructions. this has a useful effect regardless of whether caching is enabled.

It is not alpha, we could remove that annotation.

It's not really enabled by default — in fact, it's disabled unconditionally in bootsnap/setup. The intention with the readme is to encourage people to either require bootsnap/setup or think critically about the individual features and copy in the expanded form with their own opinions. In this sense, it doesn't make much sense to me to leave a feature disabled in that form.

smarquez1:
Just to add more information. I was having issues running `binding.pry`, the server would stop at an incorrect context. I fixed it by disabling `disable_trace`:
```
env = ENV['RAILS_ENV'] || "development"

Bootsnap.setup(
  cache_dir:            'tmp/cache',
  development_mode:     env == 'development',
  load_path_cache:      true,
  autoload_paths_cache: true,
  disable_trace:        env == 'production',
  compile_cache_iseq:   !ENV["COVERAGE"] ,
  compile_cache_yaml:   true
)
```

dannyfallon:
I've removed the `Alpha` annotation as it was misleading.

>It's not really enabled by default — in fact, it's disabled unconditionally in bootsnap/setup. The intention with the readme is to encourage people to either require bootsnap/setup or think critically about the individual features and copy in the expanded form with their own opinions. In this sense, it doesn't make much sense to me to leave a feature disabled in that form.

If we want to encourage people to consider each feature if they go off-piste and don't use `bootsnap/setup` then maybe `development_mode` should also be `true` in all envs for consistency? I don't think that's a good idea but I do think using `env == 'production'` is a good compromise from `true` because it minimises surprises for engineers working on large apps like ours. Thoughts?

burke:
(finally) addressed on master.

Issue #134, opened 2018-02-20
Bootsnap loads bundler from gems/ even though version in site_ruby/ is already activated


mattbrictson:
Starting with rubygems 2.7.5, Bundler is pre-installed in the `site_ruby` directory. E.g.:

```
~/.rbenv/versions/2.5.0/lib/ruby/site_ruby/2.5.0/bundler
```

You can still `gem install bundler`, at which point you will now have a second copy of Bundler here:

```
~/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1
```

Bootsnap can get confused in this scenario, leading it to load bundler again when it has already been loaded. We get lots of warnings about redeclared constants and then Bundler blows up with an error:

```
.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/constants.rb:4: warning: already initialized constant Bundler::WINDOWS
.rbenv/versions/2.5.0/lib/ruby/site_ruby/2.5.0/bundler/constants.rb:4: warning: previous definition of WINDOWS was here
.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/constants.rb:5: warning: already initialized constant Bundler::FREEBSD
.rbenv/versions/2.5.0/lib/ruby/site_ruby/2.5.0/bundler/constants.rb:5: warning: previous definition of FREEBSD was here
.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/constants.rb:6: warning: already initialized constant Bundler::NULL
[etc]
ArgumentError (Trying to register Bundler::GemfileError for status code 4 but Bundler::GemfileError is already registered)
```

---

#### Steps to reproduce

Ensure you are running rubygems 2.7.5+

```
$ gem -v
2.7.6
```

Install bundler

```
$ gem install bundler --force
```

Make a new Gemfile with `bootsnap` in it

```
$ bundle init
$ echo 'gem "bootsnap", require: false' >> Gemfile
$ bundle install
```

Run the following ruby script (e.g. in `irb`)

```ruby
require "bundler/setup" # ← this loads from site_ruby/
env = ENV["RAILS_ENV"] || "development"
require "bootsnap"
Bootsnap.setup(
  cache_dir:            ".",
  development_mode:     env == "development",
  load_path_cache:      true,
[...]

Issue #123, opened 2018-02-01
Error 'uninitialized constant Mime' in Rails 5.1/5.2


rlue:
Tried bootsnap in Rails 5.2rc1. Right out of the box, `rails s` threw this error:

```
$ rails s
/Users/rlue/.gem/ruby/2.5.0/gems/msgpack-1.2.2/lib/msgpack/version.rb:2: warning: already initialized constant MessagePack::VERSION
/Users/rlue/.config/.gem/ruby/2.5.0/gems/msgpack-1.2.2/lib/msgpack/version.rb:2: warning: previous definition of VERSION was here
/Users/rlue/.gem/ruby/2.5.0/gems/msgpack-1.2.2/lib/msgpack/msgpack.bundle: warning: already initialized constant MessagePack::ExtensionValue
/Users/rlue/.gem/ruby/2.5.0/gems/msgpack-1.2.2/lib/msgpack.rb:22: warning: already initialized constant MessagePack::DefaultFactory
/Users/rlue/.config/.gem/ruby/2.5.0/gems/msgpack-1.2.2/lib/msgpack.rb:22: warning: previous definition of DefaultFactory was here
Traceback (most recent call last):
        56: from bin/rails:3:in `<main>'
        55: from bin/rails:3:in `load'
        54: from /Users/rlue/Sandbox/foo/bin/spring:15:in `<top (required)>'
        53: from /Users/rlue/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_require.rb:70:in `require'
        52: from /Users/rlue/.rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_require.rb:70:in `require'
        51: from /Users/rlue/.config/.gem/ruby/2.5.0/gems/spring-2.0.2/lib/spring/binstub.rb:31:in `<top (required)>'
        50: from /Users/rlue/.config/.gem/ruby/2.5.0/gems/spring-2.0.2/lib/spring/binstub.rb:31:in `load'
        49: from /Users/rlue/.config/.gem/ruby/2.5.0/gems/spring-2.0.2/bin/spring:49:in `<top (required)>'
        48: from /Users/rlue/.config/.gem/ruby/2.5.0/gems/spring-2.0.2/lib/spring/client.rb:30:in `run'
[...]

Issue #122, opened 2018-01-25
superclass mismatch for class Tempfile


3wille:
Hello,

I am currently trying to upgrade our Rails ``5.1.4`` App from MRI ``1.4.1`` to MRI ``1.5.0``.
When starting the App with bootsnap, I get the following error:

```
/usr/local/opt/rbenv/versions/2.5.0/lib/ruby/2.5.0/tempfile.rb:81:in `<main>': superclass mismatch for class Tempfile (TypeError)
```

It works again after disabling ``load_path_cache``. I already tried deleting the bootsnap cache files in ``tmp/cache``. Have you any idea what might be causing this? Help would be really appreciated and thanks in advance :)

3wille:
I now found that this error does not occur on CircleCI but still on my OSX machine.

ojab:
Can you please check if https://github.com/Shopify/bootsnap/pull/136 fixes the issue? You can try it with `gem 'bootsnap', github: 'ojab/bootsnap', require: false` in Gemfile.

3wille:
Hello @ojab , when running your branch, the OSX-``system``-ruby returns a ``LoadError: No such file to load``on requiring a file that is located at ``app/lib``. The rbenv-ruby returns the same error, then segfaults.

burke:
I bet this is fixed in version 1.2.0.pre, please let me know whether that fixes this issue or not.

rafaelfranca:
@burke should not we have a tag for that release in the repository?

burke:
Yes, done.

3wille:
Hello @burke, after setting ``gem "bootsnap", "1.2.0.pre"`` in the gemfile and running ``bundle``, I still get the same error.

burke:
Can you check if 1.3.0 fixes this?

askehansen:
@burke it fixed the issue for me

Issue #114, opened 2018-01-04
rake tasks fails to run ruby v2.5.0 / rails v5.2.0.beta2


neumayr:
`rake` fails to run with latest ruby v2.5.0 / rails v5.2.0.beta2

> `already initialized constant Rake::*`

##### System 

```
$ rvm --version
rvm 1.29.3 (latest)
$ ruby -v
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin17]
$ gem --version
2.7.4
$ rake --version
rake, version 12.3.0
$ gem list | grep bootsnap
bootsnap (1.1.7)
```

<details>
<summary>Output of <code>rake -T</code></summary>

```
$ rake -T
/Users/dev/.rvm/gems/ruby-2.5.0@global/gems/rake-12.3.0/lib/rake/version.rb:3: warning: already initialized constant Rake::VERSION
/Users/dev/.rvm/rubies/ruby-2.5.0/lib/ruby/gems/2.5.0/gems/rake-12.3.0/lib/rake/version.rb:3: warning: previous definition of VERSION was here
/Users/dev/.rvm/gems/ruby-2.5.0@global/gems/rake-12.3.0/lib/rake/version.rb:6: warning: already initialized constant Rake::Version::MAJOR
/Users/dev/.rvm/rubies/ruby-2.5.0/lib/ruby/gems/2.5.0/gems/rake-12.3.0/lib/rake/version.rb:6: warning: previous definition of MAJOR was here
/Users/dev/.rvm/gems/ruby-2.5.0@global/gems/rake-12.3.0/lib/rake/version.rb:6: warning: already initialized constant Rake::Version::MINOR
/Users/dev/.rvm/rubies/ruby-2.5.0/lib/ruby/gems/2.5.0/gems/rake-12.3.0/lib/rake/version.rb:6: warning: previous definition of MINOR was here
/Users/dev/.rvm/gems/ruby-2.5.0@global/gems/rake-12.3.0/lib/rake/version.rb:6: warning: already initialized constant Rake::Version::BUILD
/Users/dev/.rvm/rubies/ruby-2.5.0/lib/ruby/gems/2.5.0/gems/rake-12.3.0/lib/rake/version.rb:6: warning: previous definition of BUILD was here
/Users/dev/.rvm/gems/ruby-2.5.0@global/gems/rake-12.3.0/lib/rake/version.rb:6: warning: already initialized constant Rake::Version::OTHER
[...]

PR #136, opened 2018-02-25
Use realpaths in Cache @path_obj

ojab:
If gems are loaded from symlinked directory, $LOAD_PATH has mixed symlinked
and real paths, while subsequent `require`s are resolved to real path, so we can
end up in situation when we rerequire the same gem with different path, `require`
loads this gem again and things go BOOM. for example:
```
From: /…/src/bootsnap/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb @ line 18 Kernel#require:

    15: def require(path)
    16:   if resolved = Bootsnap::LoadPathCache.load_path_cache.find(path)
    17:     binding.pry if path == 'set'
 => 18:     require_without_cache(resolved)
    19:   else
    20:     raise Bootsnap::LoadPathCache::CoreExt.make_load_error(path)
    21:   end
    22: rescue Bootsnap::LoadPathCache::ReturnFalse
    23:   return false
    24: rescue Bootsnap::LoadPathCache::FallbackScan
    25:   require_without_cache(path)
    26: end

[1] pry(main)> path
=> "set"
[2] pry(main)> Set.new.method(:to_s).source_location
=> ["/real_path/.rvm/rubies/ruby-2.5.0/lib/ruby/2.5.0/set.rb", 629]
[3] pry(main)> resolved
=> "/symlinked_path/.rvm/rubies/ruby-2.5.0/lib/ruby/2.5.0/set.rb"
[4] pry(main)> require_without_cache(resolved)
/symlinked_path/.rvm/rubies/ruby-2.5.0/lib/ruby/2.5.0/set.rb:625: warning: already initialized constant Set::InspectKey
/real_path/.rvm/rubies/ruby-2.5.0/lib/ruby/2.5.0/set.rb:625: warning: previous definition of InspectKey was here
=> true
[5] pry(main)> Set.new.method(:to_s).source_location
=> ["/symlinked_path/.rvm/rubies/ruby-2.5.0/lib/ruby/2.5.0/set.rb", 629]
[6] pry(main)>
```

### DIFF
--- a/lib/bootsnap/load_path_cache/cache.rb
+++ b/lib/bootsnap/load_path_cache/cache.rb
@@ -9,7 +9,7 @@ module Bootsnap
         @development_mode = development_mode
         @store = store
         @mutex = defined?(::Mutex) ? ::Mutex.new : ::Thread::Mutex.new # TODO: Remove once Ruby 2.2 support is dropped.
-        @path_obj = path_obj
+        @path_obj = path_obj.map { |f| File.exist?(f) ? File.realpath(f) : f }
         @has_relative_paths = nil
         reinitialize
       end
--- a/lib/bootsnap/load_path_cache.rb
+++ b/lib/bootsnap/load_path_cache.rb
@@ -21,12 +21,14 @@ module Bootsnap
     CACHED_EXTENSIONS = DLEXT2 ? [DOT_RB, DLEXT, DLEXT2] : [DOT_RB, DLEXT]
 
     class << self
-      attr_reader :load_path_cache, :autoload_paths_cache, :loaded_features_index
+      attr_reader :load_path_cache, :autoload_paths_cache,
+                  :loaded_features_index, :realpath_cache
 
       def setup(cache_path:, development_mode:, active_support: true)
         store = Store.new(cache_path)
 
         @loaded_features_index = LoadedFeaturesIndex.new
+        @realpath_cache = RealpathCache.new
 
         @load_path_cache = Cache.new(store, $LOAD_PATH, development_mode: development_mode)
         require_relative 'load_path_cache/core_ext/kernel_require'
@@ -53,3 +55,4 @@ require_relative 'load_path_cache/cache'
 require_relative 'load_path_cache/store'
 require_relative 'load_path_cache/change_observer'
 require_relative 'load_path_cache/loaded_features_index'
+require_relative 'load_path_cache/realpath_cache'
--- a/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
+++ b/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
@@ -36,6 +36,14 @@ module Kernel
     require_with_bootsnap_lfi(path)
   end
 
+  alias_method :require_relative_without_bootsnap, :require_relative
+  def require_relative(path)
+    realpath = Bootsnap::LoadPathCache.realpath_cache.call(
+      caller_locations(1..1).first.absolute_path, path
+    )
+    require(realpath)
+  end
+
   alias_method :load_without_bootsnap, :load
   def load(path, wrap = false)
     if resolved = Bootsnap::LoadPathCache.load_path_cache.find(path)
@@ -76,6 +84,14 @@ class << Kernel
     return false
   rescue Bootsnap::LoadPathCache::FallbackScan
     require_with_bootsnap_lfi(path)
+  end
+
+  alias_method :require_relative_without_bootsnap, :require_relative
+  def require_relative(path)
+    realpath = Bootsnap::LoadPathCache.realpath_cache.call(
+      caller_locations(1..1).first.absolute_path, path
+    )
+    require(realpath)
   end
 
   alias_method :load_without_bootsnap, :load
--- a/lib/bootsnap/load_path_cache/realpath_cache.rb
+++ b/lib/bootsnap/load_path_cache/realpath_cache.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module Bootsnap
+  module LoadPathCache
+    class RealpathCache
+      def initialize
+        @cache = Hash.new { |h, k| h[k] = realpath(*k) }
+      end
+
+      def call(*key)
+        @cache[key]
+      end
+
+      private
+
+      def realpath(caller_location, path)
+        base = File.dirname(caller_location)
+        file = find_file(File.expand_path(path, base))
+        dir = File.dirname(file)
+        File.join(dir, File.basename(file))
+      end
+
+      def find_file(name)
+        ['', *CACHED_EXTENSIONS].each do |ext|
+          filename = "#{name}#{ext}"
+          return File.realpath(filename) if File.exist?(filename)
+        end
+        name
+      end
+    end
+  end
+end
--- a/test/load_path_cache/realpath_cache_test.rb
+++ b/test/load_path_cache/realpath_cache_test.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+module Bootsnap
+  module LoadPathCache
+    class RealpathCacheTest < MiniTest::Test
+      EXTENSIONS = ['', *CACHED_EXTENSIONS]
+
+      def setup
+        @cache = RealpathCache.new
+        @base_dir = Dir.mktmpdir
+        @absolute_dir = "#{@base_dir}/absolute"
+        Dir.mkdir(@absolute_dir)
+
+        @symlinked_dir = "#{@base_dir}/symlink"
+        FileUtils.ln_s(@absolute_dir, @symlinked_dir)
+
+        real_caller = File.new("#{@absolute_dir}/real_caller.rb", 'w').path
+        symlinked_caller = "#{@absolute_dir}/symlinked_caller.rb"
+
+        FileUtils.ln_s(real_caller, symlinked_caller)
+
+        EXTENSIONS.each do |ext|
+          real_required = File.new("#{@absolute_dir}/real_required#{ext}", 'w').path
+
+          symlinked_required = "#{@absolute_dir}/symlinked_required#{ext}"
+          FileUtils.ln_s(real_required, symlinked_required)
+        end
+      end
+
+      def teardown
+        FileUtils.remove_entry(@base_dir)
+      end
+
+      def remove_required(extensions)
+        extensions.each do |ext|
+          FileUtils.rm("#{@absolute_dir}/real_required#{ext}")
+          FileUtils.rm("#{@absolute_dir}/symlinked_required#{ext}")
+        end
+      end
+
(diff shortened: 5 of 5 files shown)


ojab:
Hopefully this fixes a bunch of `already initialized constant` issues that exists today, at least it fixes rails start for me.

- I'm not sure if it's MRI bug (and `$LOAD_PATH` should always be real paths in the first place)
- I have no idea how it can be tested
- AFAIU all subsequent `require`s are resolved to real path, so we only need this on `$LOAD_PATH`, but I can be wrong 

Please advise how should I proceed with this PR.

ojab:
https://github.com/Shopify/bootsnap/issues/54
https://github.com/Shopify/bootsnap/issues/75
https://github.com/Shopify/bootsnap/issues/112
https://github.com/Shopify/bootsnap/issues/114
https://github.com/Shopify/bootsnap/issues/122
https://github.com/Shopify/bootsnap/issues/123
https://github.com/Shopify/bootsnap/issues/134

Maybe the same issue with different symptoms:
https://github.com/Shopify/bootsnap/issues/86
https://github.com/Shopify/bootsnap/issues/103
https://github.com/Shopify/bootsnap/issues/106

ojab:
And of course I'm happy to rework this PR if other solution is more suitable, but unfortunately I haven't found such solution.

burke:
The main case in which this is an issue is:

1. The user has the application source at some path on disk, say, `/Users/me/src/github.com/org/app`
2. The user symlinks the application somewhere else, say, `/Users/me/app`
4. The application adds (e.g.) `lib` to the `$LOAD_PATH`
5. The application uses a combination of `require` and `require_relative` to load files from `lib`.

For example, let's say there's a file `lib/util.rb`, which is required sometimes with `require_relative` and sometimes with `require`.

One file might require `util.rb` as `require_relative '../lib/util.rb'`; another as `require 'util'`.

If `bootsnap` is configured to resolve indexed paths to their `realpath`, then the first invocation would resolve to `/Users/me/src/github.com/org/app/lib/util.rb`, while the second would resolve to `/Users/me/app/lib/util.rb`, and the file would be double-loaded.

Here's some exploration of this issue, though it doesn't precisely reproduce the case described here:

<Details>

```ruby
require 'fileutils'

$info = ->(indent, msg) do
  STDERR.puts(format("%*s\x1b[1;34m%% \x1b[1;37m%s\x1b[0m", indent, "", msg))
end

$real_app_root      = 'realpath'
$symlinked_app_root = 'expanded_path'
$load_path_entry    = 'expanded_path/lib'
$some_file          = 'expanded_path/lib/some-file.rb'

at_exit do
  File.unlink($some_file)           rescue nil
  FileUtils.rmdir($load_path_entry) rescue nil
  File.unlink($symlinked_app_root)  rescue nil
  FileUtils.rmdir($real_app_root)   rescue nil
end

FileUtils.mkdir($real_app_root)
FileUtils.ln_s($real_app_root, $symlinked_app_root)
FileUtils.mkdir($load_path_entry)
File.write($some_file, 'puts "\x1b[1;31m!\x1b[0;31m evaluated contents\x1b[0m"')

$LOAD_PATH.unshift(File.expand_path($load_path_entry))

alias :req :require
def require(f)
  $info.(0, "require(#{f.inspect})")
  req(f)
end

def (RubyVM::InstructionSequence).load_iseq(f)
  $info.(2, "load_iseq(#{f.inspect})")
end

require 'some-file'
puts $LOADED_FEATURES.grep(/some-file/)

require_relative './expanded_path/lib/some-file'
puts $LOADED_FEATURES.grep(/some-file/)

require_relative './realpath/lib/some-file'
puts $LOADED_FEATURES.grep(/some-file/)
```

</Details>

It might be possible to work around this by also hooking `require_relative` to also `realpath` its argument.

ojab:
Idea to override `$LOAD_PATH` modifying methods and make all paths absolute on insert/whatever sounds too crazy?

ojab:
Forget it, doesn't help here.
Looks like we can't do this without `realpath` overhead on each `require_relative` :/
I'll try to think more about it and write some code this weekend.

burke:
I think it might not be *completely* crazy to add that overhead to `require_relative`.

One other possibility would be to take the `realpath` of the directory containing the target of `require_relative` and cache that, joining the basename to the resolved/cached dirname. If we wanted to be extra safe, we could stat the target and, if the file itself is a symlink, run a full `realpath` on that file to resolve it.

burke:
Something like this might do the trick:

```ruby
require 'English'

module RealpathCache
  @cache = {}

  class << self
    def call(path)
      call_rec(File.expand_path(path))
    end

    private

    def call_rec(path)
      if found = @cache[path]
        return found
      end

      # base case: FS root
      return path if File.dirname(path) == path

      if File.lstat(path).symlink?
        link_target = File.readlink(path)
        link_base   = File.dirname(path)
        expanded    = File.expand_path(link_target, link_base)
        return @cache[path] = call_rec(expanded)
      end

      File.join(
        call_rec(File.dirname(path)),
        File.basename(path)
      )
    end
  end
end

if __FILE__ == $PROGRAM_NAME
  puts RealpathCache.call(ARGV.first)
end
```

ojab:
Sorry for delay (been lazy/other stuff).
Pushed version that should correctly handle `require_relative` w/ symlinks.
Basically it's doing the same stuff that ruby does in `require_relative` under the hood, but using real directory instead of possible symlink.

Do we care that required file itself can be a symlink?

ojab:
>If we wanted to be extra safe, we could lstat the target and, if the file itself is a symlink, run a full realpath on that file to resolve it.

ojab:
>If we wanted to be extra safe, we could lstat the target and, if the file itself is a symlink, run a full realpath on that file to resolve it.
ah, looks like we care.
But we don't have actual filename (i. e. with extension), so how can we know if `file.rb` or `file.so` is actually required? Something like `CACHED_EXTENSIONS.each { |ext| File.exist?("#{file}#{ext}" }` doesn't look ok for me ._.

rlue:
Quick question, fellas — is there any way that this is related to [this Ruby bug, **require_relative and require should be compatible with each other when symlinks are used**](https://bugs.ruby-lang.org/issues/10222)? It must not be the root cause, because it looks like fixes were merged before the release of v2.5.0, but [the revision that fixed it](https://bugs.ruby-lang.org/projects/ruby-trunk/repository/ruby_2_4/revisions/62440) sounds like it should be awful relevant:

> ### Revision 62440
> 
>     file.c: rb_check_realpath
>
>     * file.c (rb_check_realpath): returns real path which has no
>       symbolic links.  similar to rb_realpath except for returning
>       Qnil if any parts did not exist.
>
>     load.c: real path to load
>
>     * load.c (rb_construct_expanded_load_path): expand load paths to
>       real paths to get rid of duplicate loading from symbolic-linked
>       directories.  [Feature #10222]

Changes were committed to Ruby trunk September of last year ([5754f15](https://github.com/ruby/ruby/commit/5754f15975d82896c2566ca01b9ce7b3122d67ec) & [b6d3927](https://github.com/ruby/ruby/commit/b6d3927e16357408720203a949cfa8741b9ebf6c)), but I'm having trouble making out how exactly this patches the `require` and `require_relative` methods.

ojab:
We're hijacking `require`/`require_relative` to load cached versions, so we're basically doing the same thing here ourself.

ojab:
Added tests for `RealpathCache` and fixed an issue when required file is a symlink itself.
Anything else should be done to get it merged?

yskkin:
As I commented in https://github.com/Shopify/bootsnap/issues/147#issuecomment-377417626, I still have an error for my case.

ojab:
Pushed version that iterates over `['', *CACHED_EXTENSIONS]` to find required file.

@yskkin please check if it works for you.

yskkin:
It works. :tada: Thank you @ojab!

burke:
I've pushed this as 1.3.0.beta, but it broke Shopify boot. I'll try to dig into the issue.

wpolicarpo:
@ojab I have the same issue here when I try to start the application in a symlink'ed folder (like `mina` and `capistrano` do). 

@burke `1.3.0.beta` also broke the normal usage (no symlinks) for me.

ojab:
@wpolicarpo could you create a new issue with description/backtrace and cc me?

lib/bootsnap/load_path_cache/cache.rb
```diff
@@ -9,7 +9,7 @@ def initialize(store, path_obj, development_mode: false)
         @development_mode = development_mode
         @store = store
         @mutex = defined?(::Mutex) ? ::Mutex.new : ::Thread::Mutex.new # TODO: Remove once Ruby 2.2 support is dropped.
-        @path_obj = path_obj
+        @path_obj = path_obj.map { |f| File.exist?(f) ? File.realpath(f) : f }
```
burke: Ah, this is the problem! `@path_obj` needs to continue to be an actual reference to `$LOAD_PATH` or to `ActiveSupport::Dependencies.autoload_paths`, and any change to it needs to trigger a corresponding event in `ChangeObserver`.

One way to do this here would be to:

```ruby
        path_obj.map! { |f| File.exist?(f) ? File.realpath(f) : f }
        @path_obj = path_obj
```

burke: It would be best too, at this point, to either:

1. Make the `ChangeObserver` understand `map!`; or
2. Make this line blow up if the `ChangeObserver` has bound to this object`.

ojab: ouch, understood, will take a look.

ojab: AFAICS ChangeObserver [already understands `map!`](https://github.com/Shopify/bootsnap/blob/master/lib/bootsnap/load_path_cache/change_observer.rb#L44), so `.map!` and we should be fine?
Any hints how it can be reproduced locally?

ojab:
@wpolicarpo could you try `gem 'bootsnap', github: 'ojab/bootsnap', require: false`?

lib/bootsnap/load_path_cache/cache.rb
burke: Oh great then.

To reproduce, I guess:

1. load bootsnap
2. Add a new dir to the load path via, say, `$LOAD_PATH.unshift`
3. Observe that bootsnap will not load features from this path since the `ChangeObserver`  hook did not fire.

An easy way to catch this is to just log or whatever from the hooked method in `ChangeObserver`.

burke:
`1.3.0.beta2` is pushed. Problem seems fixed for me.

wpolicarpo:
@ojab it's working fine now. Thanks!

gathuku:
Am having this issue any help.
```
 Unable to load application: I18n::InvalidLocaleData: can not load translations from /usr/local/bundle/gems/responders-3.0.0/lib/responders/locales/en.yml: #<NameError: uninitialized constant Bootsnap::CompileCache::ISeq>
```

PR #138, opened 2018-03-06
Fix LOADED_FEATURES issue

burke:
cc @jules2689 @lugray 

Great success:

screen shot 2018-03-06 at 3 29 28 pm

### DIFF
--- a/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
+++ b/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
@@ -11,78 +11,104 @@ module Bootsnap
 end
 
 module Kernel
-  alias_method :require_without_cache, :require
-  def require(path)
-    if resolved = Bootsnap::LoadPathCache.load_path_cache.find(path)
-      require_without_cache(resolved)
-    else
-      raise Bootsnap::LoadPathCache::CoreExt.make_load_error(path)
+  alias_method :require_without_bootsnap, :require
+
+  # Note that require registers to $LOADED_FEATURES while load does not.
+  def require_with_bootsnap_lfi(path, resolved = nil)
+    Bootsnap::LoadPathCache.loaded_features_index.register(path, resolved) do
+      require_without_bootsnap(resolved || path)
     end
-  rescue Bootsnap::LoadPathCache::ReturnFalse
-    return false
-  rescue Bootsnap::LoadPathCache::FallbackScan
-    require_without_cache(path)
   end
 
-  alias_method :load_without_cache, :load
-  def load(path, wrap = false)
+  def require(path)
+    return false if Bootsnap::LoadPathCache.loaded_features_index.key?(path)
+
     if resolved = Bootsnap::LoadPathCache.load_path_cache.find(path)
-      load_without_cache(resolved, wrap)
-    else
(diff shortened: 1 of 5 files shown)


lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
```diff
@@ -13,72 +13,106 @@ def self.make_load_error(path)
 module Kernel
   alias_method :require_without_cache, :require
   def require(path)
+    return false if Bootsnap::LoadPathCache.loaded_features_index.key?(path)
     if resolved = Bootsnap::LoadPathCache.load_path_cache.find(path)
-      require_without_cache(resolved)
+      Bootsnap::LoadPathCache.load_path_cache.register(path, resolved) do
```
rafaelfranca: Should no it be `loaded_features_index` instead of `load_path_cache`?

burke: whoah, yeah, nice catch -- thanks

lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
```diff
@@ -13,72 +13,106 @@ def self.make_load_error(path)
...
 
 class << Kernel
   alias_method :require_without_cache, :require
   def require(path)
+    return false if Bootsnap::LoadPathCache.loaded_features_index.key?(path)
+
     if resolved = Bootsnap::LoadPathCache.load_path_cache.find(path)
-      require_without_cache(resolved)
+      Bootsnap::LoadPathCache.load_path_cache.register(path, resolved) do
+        require_without_cache(resolved)
```
jules2689: Would it make sense to make `require_without_cache` do the `Bootsnap::LoadPathCache.load_path_cache.register` and return false if loaded? Might simplify the logic and reduce the chance of calling `require_without_cache` improperly

lib/bootsnap/load_path_cache/loaded_features_index.rb
```diff
@@ -0,0 +1,93 @@
...
+
+        ret
+      end
+
+      private
+
+      STRIP_EXTENSION = /\..*?$/
+      private_constant :STRIP_EXTENSION
+
+      def strip_extension(f)
```
jules2689: Is this cheaper than `File.basename`? Does that require loading files or something? Might be worth adding a comment of why we didn't use the STDLib method

lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
burke: Yeah, I'll give that a spin.

lib/bootsnap/load_path_cache/loaded_features_index.rb
jules2689: Oh is it because we don't know what extension it will be?

burke: It's because I'd actually want `File.join(File.dirname(f), File.basename(f))` which feels dumb.

jules2689: Oh, right. :/

PR #150, opened 2018-04-05
Make `path_obj` & `@path_obj` the same object again

ojab:
Fixes 1cd87f36013181b0ca6ed87b81784819dcbd0dd3


ojab:
The only test that comes to my mind is https://github.com/ojab/bootsnap/commit/938219cceee8432d10c4d3e01dbca2a16d6e5547 but I'm not sure if it's actually useful/will not break in future unrelated changes.

PR #100, opened 2017-11-30
Travis: Switch to Linux, use preinstalled ruby-2.4.1

olleolleolle:
This PR is about trying to get **faster test feedback**, by running on default nodes, which have a few versions of Ruby pre-installed. 

This build was 32s, other PRs were about 1 min.

Benefits:

- less time spent in Travis build queue
- less time spent in the `rvm` step

Drawbacks

- a test failure (see below)

PR #152, opened 2018-04-06
Fix test issues on macOS and add it to travis

ojab:
also use container-based linux

I'm not sure that macOS should be added to travis (see https://github.com/Shopify/bootsnap/pull/100), please tell if it should be removed.


burke:
Seems to work, so LGTM.

PR #160, opened 2018-04-14
Use module_function in the Kernel extension rather than code duplication

dylanahsmith:
## Problem

I noticed code duplication in the Kernel extension which has been there since the initial commit.  It appears to be there because these Kernel methods can be called on Kernel directly, which would otherwise use the ruby implementation directly.  E.g. `Kernel.require 'set'` would use the code in the `class << Kernel` block.  However, MRI ruby just uses `rb_define_global_function` to define these methods and shares the implementation, so we should be able to do the same in bootsnap.

## Solution

`rb_define_global_function` is implemented in MRI ruby using just `rb_define_module_function(rb_mKernel, name, func, argc);`, which is why there is both a private instance method and public singleton method defined for these methods.  We can do the same thing in ruby code by just using `module_function` instead of `private` without the extra method definitions explicitly on the singleton class.

### DIFF
--- a/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
+++ b/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
@@ -11,61 +11,11 @@ module Bootsnap
 end
 
 module Kernel
-  private
+  module_function
 
   alias_method :require_without_bootsnap, :require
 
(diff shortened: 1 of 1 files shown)


dylanahsmith:
The instance methods will be marked private by `module_function`, so it won't change the visibility semantics.

burke:
Perfect, :shipit:

PR #166, opened 2018-04-22
Add warning message on setup_disable_trace when ruby version is not allowed

tzmfreedom:
The `trace_instruction` option on RubyVM::InstructionSequence is removed over Ruby 2.5.0.

* https://github.com/ruby/ruby/commit/cbac40b3e53efce7740899c77eb979008dc96ff5
* https://bugs.ruby-lang.org/issues/14104

With this pullrequest, if ruby version is over 2.5.0, when the `setup_disable_trace` method called, it displays warning message, that is `this method is not allowed'

### DIFF
--- a/lib/bootsnap.rb
+++ b/lib/bootsnap.rb
@@ -35,6 +35,10 @@ module Bootsnap
   end
 
   def self.setup_disable_trace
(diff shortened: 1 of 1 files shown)


jules2689:
Thanks! sorry for the delay in merging

PR #170, opened 2018-05-23
Fix broken list formatting in README.jp [ci skip]

y-yagi:
It seems that the list is not displayed correctly because non-breaking space is included.


burke:
Thanks!

PR #175, opened 2018-06-20
Fix symlinks inside subdirectories not followed

takonomura:
Symbolic links directly under path have been followed. But symbolic links inside subdirectories have not been followed.

When the structure is:

```
├── path
│   ├── a -> ../support/a
│   ├── c
│   │   └── d -> ../../support/c/d
│   └── f
│       └── g
│           └── h.rb
└── support
    ├── a
    │   └── b.rb
    └── c
        └── d
            └── e.rb
```

Current glob does not include `path/c/d/e.rb`.

```ruby
Dir.glob("path/{,*/**}/*.rb")
#=> ["path/a/b.rb", "path/f/g/h.rb"]
```

So, Let glob follow symbolic link like below:

```ruby
Dir.glob("path/**/{,*/**}/*.rb")
#=> ["path/f/g/h.rb", "path/a/b.rb", "path/c/d/e.rb", "path/f/g/h.rb"]
```

This will make some duplicated entries, so we need to make it unique.

```ruby
Dir.glob("path/**/{,*/**}/*.rb").uniq
#=> ["path/f/g/h.rb", "path/a/b.rb", "path/c/d/e.rb"]
```

### DIFF
--- a/lib/bootsnap/load_path_cache/path_scanner.rb
+++ b/lib/bootsnap/load_path_cache/path_scanner.rb
@@ -6,13 +6,13 @@ module Bootsnap
       # Glob pattern to find requirable files and subdirectories in given path.
       # It expands to:
       #
-      #   * `/*{.rb,.so,/}` - It matches requirable files, directories and
-      #     symlinks to directories at given path.
-      #   * `/*/**/*{.rb,.so,/}` - It matches requirable files and
+      #   * `/**/*{.rb,.so,/}` - It matches requirable files, directories and
+      #     symlinks to directories at given path any directory tree depth.
+      #   * `/**/*/**/*{.rb,.so,/}` - It matches requirable files and
       #     subdirectories in any (even symlinked) directory at given path at
       #     any directory tree depth.
       #
-      REQUIRABLES_AND_DIRS = "/{,*/**/}*{#{DOT_RB},#{DL_EXTENSIONS.join(',')},/}"
+      REQUIRABLES_AND_DIRS = "/**/{,*/**/}*{#{DOT_RB},#{DL_EXTENSIONS.join(',')},/}"
       NORMALIZE_NATIVE_EXTENSIONS = !DL_EXTENSIONS.include?(LoadPathCache::DOT_SO)
       ALTERNATIVE_NATIVE_EXTENSIONS_PATTERN = /\.(o|bundle|dylib)\z/
       BUNDLE_PATH = Bootsnap.bundler? ?
@@ -45,7 +45,7 @@ module Bootsnap
           end
         end
 
-        [requirables, dirs]
+        [requirables.uniq, dirs.uniq]
       end
     end
   end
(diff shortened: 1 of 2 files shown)


takonomura:
Sorry, previous solution is not working on Ruby 2.4. It seems `path/**/{,*/**/}*/` does not match to some directories and symbolic links on Ruby 2.4.
I fixed by ac0ab4f49aead814f7745fe64b28629b29bc4750.

lib/bootsnap/load_path_cache/path_scanner.rb
```diff
@@ -34,18 +26,18 @@ def self.call(path)
...
-          else
+          if File.directory?(absolute_path)
+            dirs << relative_path
+          elsif REQUIRABLE_EXTENSIONS.include?(File.extname(relative_path))
             requirables << relative_path
           end
         end
 
-        [requirables, dirs]
+        [requirables.uniq, dirs.uniq]
```
jules2689: Im a bit concerned by the `uniq` as I assume a unique function on an array would be slow.
Otherwise, the rest of the changes seem to make sense to me.

I'd like @burke to look as well though

burke:
I'm taking a look now. I'm not sure I'm comfortable with calling `File.directory?` in the inner loop as I assume it'll do a `stat` call but I'm trying a few things out.

burke:
On the other hand, the OS will have the stat structure cached already since it was just accessed by the glob.

burke:
Ok, I think I can deal with it matching each file and redundantly calling `File.directory?`, but I'd like to see this done without having to call `uniq`, which I think we can do by doing something like:

```ruby
      ALL_TOPLEVEL_CHILDREN = "/*"
      ALL_NESTED_CHILDREN   = "/**/*/**/*"
      GLOBS                 = [ALL_TOPLEVEL_CHILDREN, ALL_NESTED_CHILDREN]
```

and

```ruby
      def self.each_file(path)
        GLOBS.each do |pattern|
          Dir.glob(path + pattern).each do |absolute_path|
            yield absolute_path
          end
        end
      end
```

What do you guys think?

lib/bootsnap/load_path_cache/path_scanner.rb
eregon: FWIW `Array#uniq` is O(n) (and not O(n^2)) as it just uses a Hash internally to deduplicate.
So I'd guess it's not a bottleneck with a Dir.glob call around.

takonomura:
Thanks. I changed glob to avoid calling `uniq`.

burke:
Oh cool, I couldn't figure out how to do that. This seems like a good fix. Thanks for digging in!

takonomura:
Thank you for merging. Do you have any plan to cut a release?

jules2689:
Yes, I think we can manage a release.

It will likely be next week since we are out of office today, and will likely be a pre-release to allow us to test everything.

The release process is a bit of extra work since we need to test with our core application to make sure that works, and allow others to do the same with the prerelease.

burke:
1.3.1 is released with this patch

PR #181, opened 2018-07-31
Display specific method name with file and line no

zoras:
'This method is not allowed with this Ruby version' is vague which really does not imply which method.


burke:
Yup, good idea. Thanks!

Issue #124, opened 2018-02-06
Bootsnap changes YAML.load_file's accepted args


dannyfallon:
:wave: The Ruby docs for `YAML.load_file` specifies a `filepath` as the argument. This then calls `File.open(filepath)` and calls `YAML.load` on the file object. However, `filepath` is not required to be a string -  you can also pass in a File Object instead of a path and it works perfectly fine in a non-Bootsnap codebase:

```ruby
> f = File.open('/tmp/test.yml')
=> #<File:/tmp/test.yml>
> YAML.load_file('/tmp/test.yml')
=> {"environment"=>"test"}
> YAML.load_file(f)
=> {"environment"=>"test"}
```

But with Bootsnap enabled:

```ruby
> f = File.open('/tmp/test.yml')
=> #<File:/tmp/test.yml>
> YAML.load_file('/tmp/test.yml')
=> {"environment"=>"test"}
> YAML.load_file(f)
Errno::ENOENT: No such file or directory - bs_fetch:open_current_file:open
from /Users/danny/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/bootsnap-1.1.8/lib/bootsnap/compile_cache/yaml.rb:50:in `fetch'
```

When we overload `YAML.load_file` in [`Bootsnap::CompileCache::YAML`](https://github.com/Shopify/bootsnap/blob/master/lib/bootsnap/compile_cache/yaml.rb#L51), we assume `path` is a string and pass `path.to_s` down to `Bootsnap::CompileCache::Native.fetch`. Calling `path.to_s` on a File object give an invalid file path, hence the error.

```ruby
> f.to_s
=> "#<File:0x007fec1dddf980>"
```

We've already updated our code to use a string into `YAML.load_file` to unblock ourselves but the behaviour was surprising enough that I'm willing to try to fix it. Would you be interested in a PR that uses `File.path` instead of `path.to_s` for calling down to `Bootsnap::CompileCache::Native.fetch` when a file object is passed in `YAML.load_file`? This would restore the behaviour we were seeing before installing Bootsnap

stevenharman:
We stumbled on this as well. We have a spec support module that we use to make `I18n` translations available for isolated specs - where Rails is not loaded:

```ruby
module I18nHelpers
  def self.included(_mod)
    locale_dir = Pathname.new(__dir__) + '../../config/locales/'
    [:en].each do |locale|
      binding.pry
[...]

PR #186, opened 2018-08-08
let YAML.load_path accept File objects

fursich:
Hi, thank you for this awesome gem!

After using bootsnap (which makes our app booting AMAZINGLY fast 😃), we realized that the interface of YAML.load_file gets slightly different from what [ruby provides](https://github.com/ruby/ruby/blob/4e0a512972cdcbfcd5279f1a2a81ba342ed75b6e/spec/ruby/optional/capi/file_spec.rb#L71-L87):

## reproduction case

For instance, the below error occurs only when we set `compile_cache_yaml` option true.

```ruby
> YAML.load_file File.new('/path/to/sample.yml')
Errno::ENOENT: No such file or directory - bs_fetch:open_current_file:open
from /my-home-directory/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bootsnap-1.3.1/lib/bootsnap/compile_cache/yaml.rb:49:in `fetch'
```

This PR is intended at fixing it.

* Related issues and discussions are also found [here](https://github.com/Shopify/bootsnap/issues/124)

## where the difference comes from

With ruby, (if I understand correctly) non-string path objects are firstly challenged with `to_path` method, and falls back to to_s (to_sym) - starts [here](https://github.com/ruby/ruby/blob/4e0a512972cdcbfcd5279f1a2a81ba342ed75b6e/include/ruby/ruby.h#L595-L596) upto [here](https://github.com/ruby/ruby/blob/4e0a512972cdcbfcd5279f1a2a81ba342ed75b6e/file.c#L172-L188)

WIth bootsnap `to_s` is applied for arguments, so when we put an arbitrary object (like a File instance above), it could be interpreted as something like "#<File:0x00007fa040cc57d0>" before passed to `bs_rb_fetch`, and thus fails when being loaded.

Hope this contribution makes senses.
(and again, a big thanks for providing this fantastic mechanism to us!)

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -269,6 +269,8 @@ cache_key_equal(struct bs_cache_key * k1, struct bs_cache_key * k2)
 static VALUE
 bs_rb_fetch(VALUE self, VALUE cachedir_v, VALUE path_v, VALUE handler)
 {
+  FilePathValue(path_v);
+
   Check_Type(cachedir_v, T_STRING);
   Check_Type(path_v, T_STRING);
 
--- a/lib/bootsnap/compile_cache/yaml.rb
+++ b/lib/bootsnap/compile_cache/yaml.rb
@@ -48,7 +48,7 @@ module Bootsnap
         klass.send(:define_method, :load_file) do |path|
           Bootsnap::CompileCache::Native.fetch(
             cache_dir,
-            path.to_s,
+            path,
             Bootsnap::CompileCache::YAML
           )
         end


ext/bootsnap/bootsnap.c
```diff
@@ -269,23 +269,25 @@ cache_key_equal(struct bs_cache_key * k1, struct bs_cache_key * k2)
 static VALUE
 bs_rb_fetch(VALUE self, VALUE cachedir_v, VALUE path_v, VALUE handler)
 {
+  VALUE stringified_path_v = rb_get_path(path_v);
```
rafaelfranca: I think we should use the `FilePathValue` macro here to avoid the value to be GCed before we use and give a segfault.

fursich: @rafaelfranca 

Sorry that was mistake.
Just replaced it with FilePathValue. ( looks much cleaner now 👍)
Thank you really a lot for the review!

fursich:
Thanks to your advise @rafaelfranca, now it's boiled down to simplest possible 😃 

Also, since it consists of just two lines' of code, I squashed the commits altogether beforehand for the sake of convenience (sorry to have cut corners.. hope it makes sense).

fursich:
Thank you @burke for the quick merge 🎉

SzNagyMisu:
I've run into this issue recently. I tried to do `rails db:migrate` and got this exception mentioned above (`Errno::ENOENT: No such file or directory - bs_fetch:open_current_file:open`).
At long last it turned out I was missing some gitignored `yml` file in the `config` folder, but I didn't get the proper error message (even after upgrading to `bootsnap 1.3.2`) until I disabled `bootsnap` in my app.

PR #193, opened 2018-09-04
Reuse expanded_path when generating index

jhawthorn:
Previously we were calling `Path#expanded_path` (which calls `File.expand_path`) for each entry we would put into the index. This would expand and create a copy of the path string for each file.

This commit stores the expanded path in a local, and reuses it for each entry in the index `Hash`. This should be faster and use less memory.


burke:
Wow, nice catch. Thanks!

PR #194, opened 2018-09-04
Add special case for uniq! in ChangeObserver

jhawthorn:
`uniq!` does not change the effective load order. The first occurrence of each path is kept, with order preserved. So we shouldn't need to reinitialize after calling it.

This is desirable because `$LOAD_PATH.uniq!` is [called by Rails](https://github.com/rails/rails/blob/5bd4a4c78b43f25245d3597876f541d84782ecc7/railties/lib/rails/engine.rb#L559).

### DIFF
--- a/lib/bootsnap/load_path_cache/change_observer.rb
+++ b/lib/bootsnap/load_path_cache/change_observer.rb
@@ -25,6 +25,14 @@ module Bootsnap
           super
         end
 
+        # uniq! keeps the first occurance of each path, otherwise preserving
+        # order, preserving the effective load path
+        def uniq!
+          ret = super
+          @lpc_observer.reinitialize if block_given?
+          ret
+        end
+
         # For each method that modifies the array more aggressively, override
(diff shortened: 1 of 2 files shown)


lib/bootsnap/load_path_cache/change_observer.rb
```diff
@@ -25,6 +25,12 @@ def concat(entries)
           super
         end
 
+        # uniq! keeps the first occurance of each path, otherwise preserving
+        # order, preserving the effective load path
+        def uniq!(&block)
```
rafaelfranca: To be safer better to just define the method as `uniq(*)` and just call `super` without passing anything. This way if future versions of ruby change the method signature we don't need to change this one.

jhawthorn: I realized that what I have is probably incorrect if `uniq!` is passed a block (It should reinitialize in that case).

PR #196, opened 2018-09-06
Avoid exponential retries of load_missing_constant on a NameError

### LINKED ISSUE #142: Object.const_get('SomeUndefinedConstant') get's really slow when using bootsnap
I have a method in my quite large rails project that checks if a class is defined like so

```ruby
def class_defined? class_name
  !!Object.const_get(class_name)
rescue NameError
  false
end
```

After installing bootsnap the method takes around 500ms to execute when I provide a class name that is not defined like: `Some::Modules::That::I::Have::SomeClassIDontHave`. I can confirm using bootsnap is the root of the slow down. Yes it might be that it's strange why I have this code, however consider `constantize` and `safe_constantize` rails methods. Those use the same kind of logic and those are slow too.

My rails project is running rails 4.2. Ruby 2.4.0

jhawthorn:
This should fix #142

The underlying load_missing_constant implementation will call itself recursively (through const_missing). If the constant can't be loaded at all, the retries will make this exponential - `O(2 ** module_depth)` 😱.

This PR makes it linear, now calling `search_for_file` only 2x as much as without bootsnap. We do this by setting a thread-local variable inside load_missing_constant to disable retries. This variable is unset in `require_or_load` so that missing constants being loaded from the required file are again attempted cached.

## Calls to search_for_file

``` ruby
ActiveSupport::Dependencies.singleton_class.prepend Module.new {
  def search_for_file(path)
    puts path
    super
  end
}

module Quux; module Quuux; module Quuuux; end; end; end

Quux::Quuux::Quuuux::SomeClassIDontHave
```

### Before (30 Calls)

```
quux/quuux/quuuux/some_class_i_dont_have
quux/quuux/some_class_i_dont_have
quux/some_class_i_dont_have
some_class_i_dont_have
some_class_i_dont_have
quux/some_class_i_dont_have
some_class_i_dont_have
some_class_i_dont_have
quux/quuux/some_class_i_dont_have
quux/some_class_i_dont_have
some_class_i_dont_have
some_class_i_dont_have
quux/some_class_i_dont_have
some_class_i_dont_have
some_class_i_dont_have
quux/quuux/quuuux/some_class_i_dont_have
quux/quuux/some_class_i_dont_have
quux/some_class_i_dont_have
some_class_i_dont_have
some_class_i_dont_have
quux/some_class_i_dont_have
some_class_i_dont_have
some_class_i_dont_have
quux/quuux/some_class_i_dont_have
quux/some_class_i_dont_have
some_class_i_dont_have
some_class_i_dont_have
quux/some_class_i_dont_have
some_class_i_dont_have
some_class_i_dont_have
```

### After (8 calls)
```
quux/quuux/quuuux/some_class_i_dont_have
quux/quuux/some_class_i_dont_have
quux/some_class_i_dont_have
some_class_i_dont_have
quux/quuux/quuuux/some_class_i_dont_have
quux/quuux/some_class_i_dont_have
quux/some_class_i_dont_have
some_class_i_dont_have
```

## 📈 Graph!

I can't help but make a graph!

Module nesting | no bootsnap | old | no nested retry | only top level retry
-- | -- | -- | -- | --
0 | 1 | 2 | 2 | 2
1 | 2 | 6 | 5 | 4
2 | 3 | 14 | 9 | 6
3 | 4 | 30 | 14 | 8
4 | 5 | 62 | 20 | 10
5 | 6 | 126 | 27 | 12
6 | 7 | 254 | 35 | 14
7 | 8 | 510 | 44 | 16
8 | 9 | 1022 | 54 | 18
9 | 10 | 2046 | 65 | 20

chart

### DIFF
--- a/lib/bootsnap/load_path_cache/core_ext/active_support.rb
+++ b/lib/bootsnap/load_path_cache/core_ext/active_support.rb
@@ -10,6 +10,14 @@ module Bootsnap
           Thread.current[:without_bootsnap_cache] = prev
         end
 
+        def self.allow_bootsnap_retry(allowed)
+          prev = Thread.current[:without_bootsnap_retry] || false
+          Thread.current[:without_bootsnap_retry] = !allowed
+          yield
+        ensure
+          Thread.current[:without_bootsnap_retry] = prev
+        end
+
         module ClassMethods
           def autoload_paths=(o)
             super
@@ -33,6 +41,12 @@ module Bootsnap
             CoreExt::ActiveSupport.without_bootsnap_cache { super }
           end
 
+          def require_or_load(*)
+            CoreExt::ActiveSupport.allow_bootsnap_retry(true) do
+              super
+            end
+          end
+
           # If we can't find a constant using the patched implementation of
           # search_for_file, try again with the default implementation.
           #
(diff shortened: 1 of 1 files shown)


rafaelfranca:
I believe the retires are there because Bootsnap's implementation of `search_for_file` only look on the bootsnap cache, so at the first try it only check the bootsnap cache, if that fail it retry without the bootsnap behavior to use the default behavior.

Issue #162, opened 2018-04-19
Bootsnap breaks loading all internal JRuby extensions


headius:
We discovered today (release day for 9.1.17.0...grumble) that bootsnap is broken on JRuby, which breaks any rails apps using it. The problem appears to be that bootsnap refuses to load any files which it cannot find physically on the filesystem. JRuby ships a number of libraries within its .jar archive, so they would not be visible to bootsnap. In addition, many of our core extensions are "virtual" in that we register a loadable entry for them but they do not exist as a .jar (or .so or .dll) anywhere in the system.

The error manifests like this:

```
$ jruby   -S rails server
NameError: uninitialized constant CGI::Util::Escape
                       const_missing at org/jruby/RubyModule.java:3370
                       <module:Util> at /Users/headius/projects/jruby/lib/ruby/stdlib/cgi/util.rb:10
                         <class:CGI> at /Users/headius/projects/jruby/lib/ruby/stdlib/cgi/util.rb:9
                              <main> at /Users/headius/projects/jruby/lib/ruby/stdlib/cgi/util.rb:8
                             require at org/jruby/RubyKernel.java:956
                     block in (root) at /Users/headius/projects/jruby/lib/ruby/gems/shared/gems/bootsnap-1.3.0-java/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:1
           require_with_bootsnap_lfi at /Users/headius/projects/jruby/lib/ruby/gems/shared/gems/bootsnap-1.3.0-java/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:21
                            register at /Users/headius/projects/jruby/lib/ruby/gems/shared/gems/bootsnap-1.3.0-java/lib/bootsnap/load_path_cache/loaded_features_index.rb:65
           require_with_bootsnap_lfi at /Users/headius/projects/jruby/lib/ruby/gems/shared/gems/bootsnap-1.3.0-java/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:20
                              <main> at /Users/headius/projects/jruby/lib/ruby/gems/shared/gems/bootsnap-1.3.0-java/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:29
                             require at org/jruby/RubyKernel.java:956
```

The actual problem, hidden by cgi.rb's load of cgi/escape, is a cryptic error from bootsnap:

```
#<Class:0x4066c471>: cannot load such file -- cgi/escape
[...]

PR #200, opened 2018-09-19
Implement JRuby path caching

ChrisBr:
This is an early suggestion to implement path caching for internal libraries for JRuby. JRuby ships with several internal libraries in jar files which bootsnap can will not find on the filesystem. In this case, we should fall back to original require / load. ``JRuby::Util.internal_libraries`` returns a list of these internal libraries.

It seems that loading is now working, however, starting of a basic Rails app is still broken because Puma fails to load socket (although the socket library seems to get loaded). It fails with this error message:

```
Bundler::GemRequireError: There was an error while trying to load the gem 'puma'.
Gem Load Error is: uninitialized constant IPSocket
```

Also there is currently also a hack for some paths which include ``.java.rb`` extension, I wasn't able to figure out where bootsnap is producing these paths yet. Would be glad for some pointers that we might can get rid of this hack (``normalize_feature`` method).

I think this patch is almost working and only some minior tweaks are necessary. Would be glad for some pointers and suggestions.

Fix https://github.com/Shopify/bootsnap/issues/162

cc @headius @kares 

// Edit:
Rails version: 5.2.1 with manually enabled bootsnap
Bootsnap master (5d1b932c95e654056969d86accfda8d2b5afc735) with this patch applied

### DIFF
--- a/lib/bootsnap/load_path_cache/cache.rb
+++ b/lib/bootsnap/load_path_cache/cache.rb
@@ -22,6 +22,20 @@ module Bootsnap
         @mutex.synchronize { @dirs[dir] }
       end
 
+      if defined? JRuby::Util.internal_libraries
+        JRUBY_INTERNAL_LIBRARIES = JRuby::Util.internal_libraries.each_with_object({}) do |lib, hash|
+          basename = File.basename(lib, '.*')
+          dirname = File.dirname(lib)
+          key = dirname == '.' ? basename : File.join(dirname, basename)
+
+          hash[key] = key # jruby/util => jruby/util
+          hash[lib] = lib # jruby/util.rb => jruby/util.rb
+          hash[key + DOT_SO] = key + DOT_SO # jruby/util.so => jruby/util.so
+        end.freeze
+      else
+        JRUBY_INTERNAL_LIBRARIES = {}.freeze
+      end
+
       # { 'enumerator' => nil, 'enumerator.so' => nil, ... }
       BUILTIN_FEATURES = $LOADED_FEATURES.each_with_object({}) do |feat, features|
         # Builtin features are of the form 'enumerator.so'.
@@ -45,6 +59,11 @@ module Bootsnap
       def find(feature)
         reinitialize if (@has_relative_paths && dir_changed?) || stale?
         feature = feature.to_s
+
+        if JRUBY_INTERNAL_LIBRARIES.key?(feature) && !BUILTIN_FEATURES.key?(feature)
+          return JRUBY_INTERNAL_LIBRARIES[feature]
+        end
+
         return feature if absolute_path?(feature)
         return File.expand_path(feature) if feature.start_with?('./')
         @mutex.synchronize do
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,11 @@ os:
 rvm:
   - ruby-2.4
   - ruby-2.5
+  - jruby-9.2.0.0
+
+matrix:
+  allow_failures:
+    - rvm: jruby-9.2.0.0
 
 before_script: rake
 script: bundle exec bin/testunit


rafaelfranca:
I'm not sure if I want to support JRuby in this library.

We are not Java users so it would have to rely on people that use Java to make sure things works and are being evolved correctly.

Even just setting CI to use JRuby is not a very pleasant situation, at least not in travis-ci. The build fails for no reason most of times and often require us to lost hours trying to understand the reason or just ignore JRuby builds that often means that the library will stop working on JRuby at one time.

This diff looks harmless though and I'm fine to apply it if the other maintainers agree. But saying that this library supports JRuby is something that I really want to avoid. For the reasons described above.

lib/bootsnap/load_path_cache/cache.rb
```diff
@@ -40,11 +51,23 @@ def load_dir(dir)
         end
       end.freeze
 
+      def normalize_feature(feature)
```
rafaelfranca: We should also conditionally define this method. I don't think MRI users should need to do all this calculation just to return false in `JRUBY_INTERNAL_LIBRARIES.key?(normalize_feature(feature))`

kares:
Hey Chris, could you maybe also try setting up JRuby CI with `allow_failures: ` just to see if it behaves non-reliably. most folks just setup testing against JRuby on CI and forget about it ...

ChrisBr:
> Hey Chris, could you maybe also try setting up JRuby CI with allow_failures: just to see if it behaves non-reliably. most folks just setup testing against JRuby on CI and forget about it ..

Done!

lib/bootsnap/load_path_cache/cache.rb
```diff
@@ -22,6 +22,17 @@ def load_dir(dir)
         @mutex.synchronize { @dirs[dir] }
       end
 
+      if defined? JRuby::Util.internal_libraries
```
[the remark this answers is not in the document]
ChrisBr: The difference is that BUILTIN_FEATURES for MRI return false for require and therefore bootsnap seems to ignore it (``raise LoadPathCache::ReturnFalse``). In JRuby case we still need to load the internal library in this case.

But I'm not too sure about these internals of bootsnap, maybe somebody of the maintainers can clarify?

lib/bootsnap/load_path_cache/cache.rb
ChrisBr: I do now all the normalization while creating the hash. So in MRI case it is just an empty hash. So no conditional needed anymore or what do you think @rafaelfranca ?

ChrisBr:
@kares this seems now to work for me mostly, could you maybe double check?

I still get an error because bootsnap tries to load date.rb from the rails root directory, not sure what is going on there? Otherwise it works including puma starts now as well.

headius:
@rafaelfranca I would like to know what issues you ran into running JRuby on Travis. I am confused how you have managed to do so since this library has up to now been specific to MRI and there's no way tests would have passed, if they even ran.

In any case...Rails depends on this library now, and since Rails supports JRuby, we are trying to make sure all dependent libraries at least *load* on JRuby even if they don't do 100% of what they might do on MRI. Perhaps you should lobby to remove Bootsnap as a Rails dependency if you would prefer to only run on MRI?

We are just trying to keep Rails working for JRuby users. Every time a new dependency is added we have to go through this, or else abandon all the JRuby shops out there. We would appreciate a bit of help, but we are also willing to do the extra work necessary if our changes can be included in a release.

rafaelfranca:
> @rafaelfranca I would like to know what issues you ran into running JRuby on Travis.

This gem never had an issue on JRuby because we don't support it, but most of the gems that I maintain that used to have JRuby in CI had broken builds with JRuby to the point that was better to the gem to remove support. Rails included. Some examples:

https://github.com/erikhuda/thor/commit/264d87c2ebec7ff46fe9cbeab27fd16b0abfbc1d
https://travis-ci.org/rack/rack/builds/427963386
https://github.com/plataformatec/devise/commit/611e0335cc35ad6b0c148c543531c5daa7dbdd90#diff-354f30a63fb0907d4ad57269548329e3
https://travis-ci.org/rails/rails/jobs/431737655

> Rails supports JRuby

Does it? I'm not aware of it. The Rails test suite don't pass on JRuby and the Rails core team do not [officially support JRuby since July 2015](https://github.com/rails/rails/commit/089ad5fd358333c8dc4971b965688e28ce7e1dd4).

> Perhaps you should lobby to remove Bootsnap as a Rails dependency if you would prefer to only run on MRI?

I think Rails applications [running on JRuby don't use bootsnap](https://github.com/rails/rails/commit/b10f371366a606310cab26648d798836e030bdc8). Is not that sufficient?

> We would appreciate a bit of help, but we are also willing to do the extra work necessary if our changes can be included in a release.

I do understand the issue but I'm not sure how to fix it. Like I said because we are not JRuby users it is really hard to maintain JRuby support in the libraries we create. When issues happens it is easier to remove support or let the test suite broken than trying to fix issues that are specific for JRuby and we have no knowledge how to debug and fix.

I think one solution here would be to not include bootsnap for Ruby applications and Rails is already doing that.

ChrisBr:
So does it make sense to continue working on this one or should we just close it then?

cc @rafaelfranca @headius

rafaelfranca:
For now, let's finish this. We can revisit this if something is broken in the future and it became maintenance burden.

headius:
I am going to modify builtin libraries to look like regular libraries in JRuby 9.2.6, so I think this is unnecessary now. Also going to deprecate that `internal_libraries` method since it doesn't mean anything now.

cc @enebo @kares @ChrisBr @rafaelfranca

ChrisBr:
@headius thanks for the update, so we' going to close this PR?

headius:
@ChrisBr Yes, this PR should be unnecessary. Bootsnap on JRuby 9.2.6+ should work correctly as-is.

headius:
This is up to you, but if you want to add a JRuby version check, that would help out any folks that can't upgrade JRuby but need to use Rails 5.2.x (whichever one had bootsnap enabled by default) or if they are migrating some other project that uses bootsnap.

PR #202, opened 2018-09-21
Add bundler? module test

kazu9su:
Add bundler? method test

### DIFF
--- a/test/bundler_test.rb
+++ b/test/bundler_test.rb
@@ -1,0 +1,7 @@
+require 'test_helper'
+
+class BundlerTest < Minitest::Test
+  def test_bundler?
+    assert Bootsnap.bundler?
+  end
+end


jules2689:
I think it makes sense to add a basic test here, but I think we need to add a bit more.
If we look at the `bundler?` method source, we have a few different avenues:
```ruby
def bundler?
    return false unless defined?(::Bundler)

    # Bundler environment variable
    ['BUNDLE_BIN_PATH', 'BUNDLE_GEMFILE'].each do |current|
      return true if ENV.key?(current)
    end

    false
end
```

We should test when `::Bundler` is not defined:
```ruby
def without_bundler
  b = ::Bundler
  begin
    Object.send(:remove_const, :Bundler)
    yield
  ensure
    Object.send(:const_set, :Bundler, b)
  end
end

test 'bundler? is false without Bundler const' do
  without_bundler do
    refute_predicate Bootsnap, :bundler?
  end
end
```

We should then do 2 more tests that set 'BUNDLE_BIN_PATH', 'BUNDLE_GEMFILE' to something and assert true, then remove them both and assert false. Make sure to return to the original value in a `begin..ensure` block

test/bundler_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class BundlerTest < Minitest::Test
+  def test_bundler?
+    assert Bootsnap.bundler?
```
jules2689: We prefer the `assert_predicate` method: `assert_predicate Bootsnap, :bundler?` because it gives a better error message.

Although we're running in a Bundler environment, I'd prefer not to rely on that to make this test work. We should explicitly set this up for success or failure

kazu9su:
@jules2689 Thanks for your kindful advice;)

I updated test cases as you mentioned. Please confirm that.
(Btw, `test do block` is a feature of [ActiveSupport::TestCase](https://guides.rubyonrails.org/testing.html#rails-meets-minitest), so I don't use at this time)

test/bundler_test.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  def test_bundler_without_bundler_const
+    without_bundler do
+      refute_predicate Bootsnap, :bundler?
+    end
+  end
+
+  def test_bundler_without_required_env_keys
```
jules2689: Doesn't this one need to be _with_ the `bundler` const for sure?

test/bundler_test.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def test_bundler_with_bundle_gemfile_env
+    without_required_env_keys do
+      ENV['BUNDLE_GEMFILE'] = 'foo'
+      assert_predicate Bootsnap, :bundler?
+    end
+  end
+
+  def without_bundler
```
jules2689: I'd prefer this be a private method

test/bundler_test.rb
```diff
@@ -0,0 +1,52 @@
+require 'test_helper'
+
+class BundlerTest < Minitest::Test
+  def without_required_env_keys
```
jules2689: This is better as a private method

test/bundler_test.rb
```diff
@@ -0,0 +1,52 @@
...
+    e = {}
```
jules2689: It would be more clear if this was `original_env` instead of `e`

test/bundler_test.rb
```diff
@@ -0,0 +1,55 @@
...
+      refute_predicate Bootsnap, :bundler?
+    end
+  end
+
+  private
+
+  def without_bundler
+    b = ::Bundler
+    begin
+      Object.send(:remove_const, :Bundler)
```
burke: I feel like there should be a cleaner way to accomplish this but I can't think of one.

test/bundler_test.rb
```diff
@@ -0,0 +1,55 @@
...
+      yield
+    ensure
+      Object.send(:const_set, :Bundler, b)
+    end
+  end
+
+  def without_required_env_keys
+    original_env = {}
+    begin
+      %w[BUNDLE_BIN_PATH BUNDLE_GEMFILE].each do |k|
```
burke: It might not be enforced here but our style guide usually asks for `%w()` instead of `%w[]`.

jules2689:
Thanks @kazu9su !!

PR #217, opened 2018-12-12
Test ruby-head with allow_failures

k0kubun:
Sometimes Ruby trunk crashes on bootsnap. 

I think it would be nice if we can notice trunk's change which is harmful for bootsnap earlier by this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,11 @@ os:
 rvm:
   - ruby-2.4
   - ruby-2.5
+  - ruby-head
+
+matrix:
+  allow_failures:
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thanks @k0kubun. This definitely help, but given the build only run when commits are made, and there is no much activity happening here it would still be hard to catch those. I'll setup a periodic daily build to see if that helps.

k0kubun:
> I'll setup a periodic daily build to see if that helps.

:+1: It would be definitely helpful at least for me when distinguishing some possible causes/revisions.

PR #221, opened 2019-01-22
Make bootsnap/setup a noop on unsupported ruby implementations

casperisfine:
### Context

We (Shopify) don't really want to maintain support for alternative implementations of Ruby because it requires quite a lot of knowledge of the platforms and have a tendency to fail often.

However bootsnap is currently a burden to them, because they often try to test their compatibility with existing Rails apps (e.g. Discourse) and bootsnap prevent from doing that without modifying the app.

### Proposal

We don't actually have to implement & test load path caching or ISeq caching on JRuby / Truffle / Rubinius, we can simply make sure Bootsnap acts as a noop on these alternative implementations.

This is what this PR attempt to do.

cc @eregon

### DIFF
--- a/test/minimal_support/bootsnap_setup.rb
+++ b/test/minimal_support/bootsnap_setup.rb
@@ -1,0 +1,2 @@
+require 'bundler/setup'
+require 'bootsnap/setup'
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,15 @@ rvm:
 matrix:
   allow_failures:
     - rvm: ruby-head
+  include:
+    - rvm: jruby
+      os: linux
+      env: MINIMAL_SUPPORT=1
+    # Truffle 1.0.0-rc12 should be able to pass:
+    # https://github.com/oracle/truffleruby/issues/1400#issuecomment-456166426
+    # It will be release early Feb 2019
+    # - rvm: truffleruby
+    #   os: linux
+    #   env: MINIMAL_SUPPORT=1
 
-before_script: rake
-script: bundle exec bin/testunit
+script: bin/ci
--- a/bin/ci
+++ b/bin/ci
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+if [[ "${MINIMAL_SUPPORT-0}" -eq 1 ]]; then
+  exec bin/test-minimal-support
+else
+  rake
+  exec bin/testunit
+fi
--- a/bin/test-minimal-support
+++ b/bin/test-minimal-support
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+cd test/minimal_support
+bundle
+BOOTSNAP_CACHE_DIR=/tmp bundle exec ruby -I ../../lib bootsnap_setup.rb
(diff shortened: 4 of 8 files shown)


test/minimal_support/bootsnap_setup.rb
```diff
@@ -0,0 +1,2 @@
+require 'bundler/setup'
+require 'bootsnap/setup'
```
eregon: Would `Bootsnap.setup` also work here?
I think that should be tested too, as it's used, e.g., in Discourse:
https://github.com/discourse/discourse/blob/3c9c95ac837b1222c2dce43294dd69912a4f43f0/config/boot.rb#L20-L29

casperisfine: Yes, ultimately `bootsnap/setup` simply call `Bootsnap.setup` with default options.

eregon:
I think this is a good approach to let the gem work on alternative implementations :+1: 

Later on, maybe some of the caching can be made to work on alternative implementations.
What would be really useful for that is to have an estimate of which type of caching gains how much of startup on, e.g., Shopify or Discourse.

Compilation caching of YARV bytecode is obviously MRI-specific, but other caching seems like it could work on alternative implementations. `YAML.load_file` caching for instance seems like it should be portable.
I guess `require`/`load`/`$LOAD_PATH` tends to have some implementation-specific logic (e.g., paths), but is otherwise probably fairly portable as well, but likely already more subtle to get right.

test/minimal_support/bootsnap_setup.rb
eregon: I see, thanks for the pointer, looks good.

casperisfine:
Yes, later support can be envisioned, but it can also turn into a huge support burden for us (see: https://github.com/Shopify/bootsnap/pull/200#issuecomment-423710621). 

But yeah, YAML and load path caches should be fairly portable.

eregon:
Is there a release planned for bootsnap, including this change?
That would be very useful for TruffleRuby (and I guess JRuby too).

casperisfine:
@eregon I checked the unreleased change, and beside this PR, it's only extremely minor fixes, so I'm confident I can release a new version even though I'm not directly a maintainer.

However I'd like to make sure it actually work with truffle. Since it seems the RC that fixes messagepack isn't released yet, did you test bootsnap master with truffle master?

eregon:
Sorry, I should have checked earlier, after the PR was merged, but unfortunately I got busy and forgot the next steps to get bootsnap running.

casperisfine:
No worries.

eregon:
Now that https://github.com/Shopify/bootsnap/pull/229 is merged, we can be confident bootsnap `master` works with TruffleRuby and JRuby.

I think it would be great to have a release, so applications using bootsnap can benefit from the improved compatibility.
Could you plan or schedule a `bootsnap` release?

casperisfine:
Yes it's planned. I just need to check with @burke wether we should publish a `1.3.3` or a `1.4.0` (IMO `1.3.3`).

He's on the east coast timezone, so just give us a couple hours.

eregon:
Sounds good :+1:

PR #226, opened 2019-02-01
Only require load_path_cache files if supported

eregon:
* This saves a bit of time on unsupported implementations and avoids loading fileutils and msgpack when not needed.
* It also works around an issue in TruffleRuby, see https://github.com/oracle/truffleruby/issues/1562

I will fix that issue, but it is too late for the upcoming release (1.0.0-rc12).
I can also confirm `MINIMAL_SUPPORT=1 bin/ci` passes on TruffleRuby with that issue fixed (even before this PR).

Relates to #221.

Do you know if `Bootsnap::ExplicitRequire.from_rubylibdir` is used outside of the single usage in `lib/bootsnap/load_path_cache/store.rb` ? It doesn't seem so from [GitHub code search](https://github.com/search?q=Bootsnap%3A%3AExplicitRequire.from_rubylibdir+-filename%3Astore.rb+-filename%3Aexplicit_require.rb&type=Code).

### DIFF
--- a/lib/bootsnap/load_path_cache.rb
+++ b/lib/bootsnap/load_path_cache.rb
@@ -59,10 +59,12 @@ module Bootsnap
   end
 end
 
-require_relative 'load_path_cache/path_scanner'
-require_relative 'load_path_cache/path'
-require_relative 'load_path_cache/cache'
-require_relative 'load_path_cache/store'
-require_relative 'load_path_cache/change_observer'
-require_relative 'load_path_cache/loaded_features_index'
-require_relative 'load_path_cache/realpath_cache'
+if Bootsnap::LoadPathCache.supported?
+  require_relative 'load_path_cache/path_scanner'
+  require_relative 'load_path_cache/path'
+  require_relative 'load_path_cache/cache'
+  require_relative 'load_path_cache/store'
+  require_relative 'load_path_cache/change_observer'
+  require_relative 'load_path_cache/loaded_features_index'
+  require_relative 'load_path_cache/realpath_cache'
+end


casperisfine:
Also to add more context. I didn't conditionally load in my PR to reduce as much as possible the difference between the different implementations.

e.g. some app depending on `msgpack` being required by Bootsnap would break if you try running it with Truffle/JRuby.

It's not a huge deal though.

eregon:
Right. I think it's not too bad to require users to be explicit in such a case, as applications should depend on and `require 'msgpack'` if they use it and not depend on Bootsnap loading it as a side effect.

In Rails apps, if `Bundler.require` is used, I think `msgpack` is loaded anyway so there it shouldn't even be observable.

casperisfine:
> I think it's not too bad to require users to be explicit in such a case, as applications should depend on and `require 'msgpack'` if they use it and not depend on Bootsnap loading it as a side effect.

Definitely no big deal. I just figured we might as well have as little difference as possible. 

> In Rails apps, if Bundler.require is used, I think msgpack is loaded anyway so there it shouldn't even be observable.

Only if messagepack is directly reference in your Gemfile. If it's only there as a bootsnap dependency, Bundler won't require it automatically.

eregon:
> I just figured we might as well have as little difference as possible.

`Bootsnap::CompileCache` already loads the files only if supported, so as a bonus the change makes it consistent between LoadPathCache and CompileCache.

> Only if messagepack is directly reference in your Gemfile. If it's only there as a bootsnap dependency, Bundler won't require it automatically.

Makes sense, I wasn't sure and you are definitely more knowledgeable about that :)

eregon:
Thanks for merging!

PR #230, opened 2019-02-05
Respond to $LOADED_FEATURES.delete by removing from LFI

burke:
@fxn @rafaelfranca @kaspth this seems to solve the problem, WDYT?

```
bootsnap deletion-lfi > irb -rbootsnap/setup
irb(main):001:0> Bootsnap::LoadPathCache.loaded_features_index.key?('bootsnap')
=> true
irb(main):002:0> x=$LOADED_FEATURES.detect{|x|x=~/bootsnap.rb/}
=> "/Users/burke/.gem/ruby/2.6.0/gems/bootsnap-1.3.2/lib/bootsnap.rb"
irb(main):003:0> $LOADED_FEATURES.delete(x)
=> "/Users/burke/.gem/ruby/2.6.0/gems/bootsnap-1.3.2/lib/bootsnap.rb"
irb(main):004:0> Bootsnap::LoadPathCache.loaded_features_index.key?('bootsnap')
=> false
irb(main):005:0>
```

### DIFF
--- a/lib/bootsnap/load_path_cache.rb
+++ b/lib/bootsnap/load_path_cache.rb
@@ -37,6 +37,7 @@ module Bootsnap
 
         @load_path_cache = Cache.new(store, $LOAD_PATH, development_mode: development_mode)
         require_relative('load_path_cache/core_ext/kernel_require')
+        require_relative('load_path_cache/core_ext/loaded_features')
 
         if active_support
           # this should happen after setting up the initial cache because it
--- a/lib/bootsnap/load_path_cache/cache.rb
+++ b/lib/bootsnap/load_path_cache/cache.rb
@@ -73,7 +73,8 @@ module Bootsnap
             x = search_index(feature[0..-4] + DLEXT)
             return x if x
             if DLEXT2
-              search_index(feature[0..-4] + DLEXT2)
+              x = search_index(feature[0..-4] + DLEXT2)
+              return x if x
             end
           else
             # other, unknown extension. For example, `.rake`. Since we haven't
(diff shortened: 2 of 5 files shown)


fxn:
@burke I believe this might work. Let me try playing a bit, I'll write back.

fxn:
Let me first understand this failure:

```ruby
$:.unshift("foo")

dirs = %w(foo cache)
FileUtils.rm_rf(dirs)
FileUtils.mkdir(dirs)

module Kernel
  alias_method :original_test_require, :require

  def require(path)
    puts "calling Kernel#require"
    original_test_require(path).tap do
      puts "loaded #{$LOADED_FEATURES.last}"
    end
  end
end

require "bootsnap"
Bootsnap.setup(
  cache_dir: "cache",
  autoload_paths_cache: false
)

File.write("foo/x.rb", "X = :FOO")

require "x"
# `require': cannot load such file -- x (LoadError)
```

Why does it fail? Would it file in the context of a Rails application in which a user adds a file and reloads (perhaps the AS monkey patches address this)?

burke:
I added another commit to this PR that will make bootsnap always scan the `$LOAD_PATH` on a cache miss when `development_mode: true` is passed to `Bootsnap.setup`.

fxn:
Hey! There are more questions regarding this patch!

Next one is:

```ruby
require "fileutils"

$:.unshift("foo")

dirs = %w(foo cache)
FileUtils.rm_rf(dirs)
FileUtils.mkdir(dirs)

module Kernel
  alias_method :original_test_require, :require

  def require(path)
    puts "calling Kernel#require"
    original_test_require(path).tap do
      puts "loaded #{$LOADED_FEATURES.last}"
    end
  end
end

require "bootsnap"
Bootsnap.setup(
  cache_dir: "cache",
  autoload_paths_cache: false
)

x_rb = File.expand_path("foo/x.rb", __dir__)

File.write(x_rb, "X = :FOO")
require "x"
p X

Object.send(:remove_const, :X)
$LOADED_FEATURES.delete(x_rb)
require "x"
p X
```

The last `require` does not call `Kernel#require`, and the following `p X` errs with `NameError`, is that correct?

burke:
huh, yeah, crap, that doesn't seem to work. I'll keep poking at it then, thanks.

PR #232, opened 2019-02-06
Try harder to figure out which feature was added when not pre-resolved.

burke:
cc @fxn, will release as 1.4.0.pre3 immediately


fxn:
Thanks @burke! It's late here in Barcelona, but tomorrow I'll check things out.

Don't worry about releases (at least as far as these changes is concerned), I'll test things against `master` for a while.

PR #237, opened 2019-02-21
add dependency: rb-readline

brookzhang:
To fix this error:
/usr/local/lib/ruby/gems/2.5.0/gems/bootsnap-1.3.2/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require': cannot load such file -- readline (LoadError)


rafaelfranca:
Thank you for the pull request but bootsnap doesn't depend on readline. You are seeing bootsnap in the backtrace but this error is likely caused by other library or your application that is requiring readline.

PR #248, opened 2019-03-19
expand_path correctly when resolving relative paths:

burke:
Without this change, this code:

    autoload(:B, './b')

Resolves to a path:

    $PWD/b # for example, /Users/me/src/foo/bar/app/lib/b

This is missing the .rb extension, which is normally fine because ruby
is happy to resolve it when we pass it to require, etc., but we also
insert that resolved path into our LoadedFeaturesIndex. Later, if we
attempt to purge that file from the LFI, we will generally have the full
path available, and therefore remove a non-existent entry, leaving the
not-fully-expanded one.

cc @Edouard-chin @byroot

### DIFF
--- a/lib/bootsnap/load_path_cache/cache.rb
+++ b/lib/bootsnap/load_path_cache/cache.rb
@@ -46,7 +46,7 @@ module Bootsnap
         reinitialize if (@has_relative_paths && dir_changed?) || stale?
         feature = feature.to_s
         return feature if absolute_path?(feature)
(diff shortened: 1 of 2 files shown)


Edouard-chin:
Confirm it works perfectly. The repro script is now passing https://gist.github.com/Edouard-chin/f3092baab8544d97c91a9c3e3da70c8d and I pointed Shopify on this branch.

Really appreciate your help Burke ❤️ , I was lost on how to fix it.

PR #252, opened 2019-03-25
Add a hook for $LOADED_FEATURES.reject!

casperisfine:
Ref: https://github.com/fxn/zeitwerk/pull/32

TL;DR; Zeitwerk is removing paths one by one using `$LOADED_FEATURES.delete`, but that's terribly slow (`O(n*m)` more or less).

We'd like to use `$LOADED_FEATURES.reject!` instead to improve performance (over 30 times faster for Shopify core, but the bigger the `$LOADED_FEATURES` is, the slower it will be, so it might not be as noticeable on smaller apps).

From my initial test this is somewhat performant enough. My test scenario is unloading `172` paths out of `7585`, and it takes about 200ms (1s with Zeitwerk + Bootsnap).

I suppose this could be optimized further as the comments suggest, but I believe this PR is enough to change Zeitwerk reloading from "too slow to even be usable" to "could be better".

cc @fxn

### DIFF
--- a/lib/bootsnap/load_path_cache/core_ext/loaded_features.rb
+++ b/lib/bootsnap/load_path_cache/core_ext/loaded_features.rb
@@ -4,4 +4,14 @@ class << $LOADED_FEATURES
     Bootsnap::LoadPathCache.loaded_features_index.purge(key)
     delete_without_bootsnap(key)
   end
+
+  alias_method(:reject_without_bootsnap!, :reject!)
+  def reject!(&block)
+    backup = dup
+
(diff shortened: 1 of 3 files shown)


casperisfine:
@burke I'll merge that tomorrow. Mind if I release a new version as well ? Or is there other stuff you'd like included ?

casperisfine:
I released this as `1.4.2.rc3`, let me know if it's good for you to release as final `1.4.2`.

burke:
Yeah, rc2 was good to release AFAICT, I just hadn't got to it :+1:

Issue #254, opened 2019-03-27
Unable to load application: Bootsnap::CompileCache::PermissionError


afuno:
I have a new application on Rails (~> 5.2.2), which I am developing the second day. That is an elementary application.

Today I decided to deploy it on a server for testing. For this I use Dokku.

I have never had a similar problem with Dokku. At the very last deployment step, when the application starts up, I get a huge log with errors. Here is the part (beginning):

```
       CHECKS file not found in container: Running simple container check...
-----> Waiting for 10 seconds ...
8b42e49d18cd9a956ea239c859a27c5906167b5a9842c0daf85817e4faed70ad
remote: App container failed to start!!
=====> st-qc web container output:
       Starting web...
       web| Puma starting in single mode...
       web| * Version 3.12.1 (ruby 2.6.2-p47), codename: Llamas in Pajamas
       web| * Min threads: 5, max threads: 5
       web| * Environment: production
       web| ! Unable to load application: Bootsnap::CompileCache::PermissionError: bootsnap doesn't have permission to write cache entries in '/app/tmp/cache/bootsnap-compile-cache' (or, less likely, doesn't have permisison to read '/app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.2/lib/bootsnap/compile_cache/yaml.rb')
       web| bundler: failed to load command: puma (/app/vendor/bundle/ruby/2.6.0/bin/puma)
       Bootsnap::CompileCache::PermissionError: bootsnap doesn't have permission to write cache entries in '/app/tmp/cache/bootsnap-compile-cache' (or, less likely, doesn't have permisison to read '/app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.2/lib/bootsnap/compile_cache/yaml.rb')
         /app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.2/lib/bootsnap/compile_cache.rb:28:in `permission_error'
         /app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.2/lib/bootsnap/compile_cache/iseq.rb:43:in `rescue in load_iseq'
         /app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.2/lib/bootsnap/compile_cache/iseq.rb:33:in `load_iseq'
         /app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.2/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:22:in `require'
[...]

Issue #171, opened 2018-05-29
Atomic write cache file error.


yarhtut:
Can not access to `bin/rails c` on linux
ruby/2.4.0/gems/bootsnap-1.3.0/lib/bootsnap/compile_cache/iseq.rb:37:in `fetch': Permission denied - bs_fetch:atomic_write_cache_file:open (Errno::EACCES)

We know that it is a permissions issue with our server user. As we deploy our application as a `deployer` user and are able to access the console using that role.

However when we atempt to run rails console as a `developer` role we get the following errors:

```
[WARNING] Could not load command "rails/commands/server/server_command". Error: Permission denied - bs_fetch:atomic_write_cache_file:open.
.../rails/shared/bundle/ruby/2.4.0/gems/bootsnap-1.3.0/lib/bootsnap/compile_cache/iseq.rb:37:in `fetch'
.../rails/shared/bundle/ruby/2.4.0/gems/bootsnap-1.3.0/lib/bootsnap/compile_cache/iseq.rb:37:in `load_iseq'
.../rails/shared/bundle/ruby/2.4.0/gems/bootsnap-1.3.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:21:in `require'
.../rails/shared/bundle/ruby/2.4.0/gems/bootsnap-1.3.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:21:in `block in require_with_bootsnap_lfi'
.../rails/shared/bundle/ruby/2.4.0/gems/bootsnap-1.3.0/lib/bootsnap/load_path_cache/loaded_features_index.rb:65:in `register'
.../rails/shared/bundle/ruby/2.4.0/gems/bootsnap-1.3.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:20:in `require_with_bootsnap_lfi'
.../rails/shared/bundle/ruby/2.4.0/gems/bootsnap-1.3.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:29:in `require'
.../rails/shared/bundle/ruby/2.4.0/gems/activesupport-5.2.0/lib/active_support/dependencies.rb:283:in `block in require'
.../rails/shared/bundle/ruby/2.4.0/gems/activesupport-5.2.0/lib/active_support/dependencies.rb:249:in `load_dependency'
.../rails/shared/bundle/ruby/2.4.0/gems/activesupport-5.2.0/lib/active_support/dependencies.rb:283:in `require'
[...]

PR #255, opened 2019-04-02
Cache permissions and umask issues after switch to mkstemp

kpumuk:
This PR resolves multiple permissions-related issues introduced in https://github.com/Shopify/bootsnap/pull/240. 

## State before the change

* When directory does not exist, it is being created with `mkpath` (respects `umask`, default permissions `0775`), and then file opened with `open` (respects `umask`, default permissions are `0664`).
* When the second file is created in the same directory, it uses `mkstemp`, which always creates files with permissions `0600`. **This causes various issues when permissions system relies on user groups, as the group can no longer read or write cache files.**

## After this change

* After directory first created, we use `open` to create first file.
* The second file is created with `mkstemp` call, which sets permissions to `0600`.
* To fix permissions, we apply `0644` to the file after rename, and properly use umask value (current `umask` is determined in the `Init_bootsnap`)

I have tested this change with and without umask, and it seems to be working properly:

* By default directories are created with `0775`
* Files are created with `0664`
* When umask `022` is specified, directories are `0755` and files are `0644`

## Previous behavior that was reverted by `mkstemp`

* https://github.com/Shopify/bootsnap/commit/be4e8476954f7c70b78a7bcf2e1fa45aa7008e77

### Related issues

* https://github.com/Shopify/bootsnap/issues/254 Unable to load application: Bootsnap::CompileCache::PermissionError
* https://github.com/Shopify/bootsnap/issues/171 Atomic write cache file error

### Updates

I was wrong about the second `open` call, it makes sense now and PR is corrected. ;-) The pemissions issue is still there.

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -74,6 +74,8 @@ static uint32_t current_ruby_platform;
 static uint32_t current_ruby_revision;
 /* Invalidates cache when RubyVM::InstructionSequence.compile_option changes */
 static uint32_t current_compile_option_crc32 = 0;
+/* Current umask */
+static mode_t current_umask;
 
 /* Bootsnap::CompileCache::{Native, Uncompilable} */
 static VALUE rb_mBootsnap;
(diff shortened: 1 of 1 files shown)


kpumuk:
Thank you!

markets:
Thanks a lot for working on this! :raised_hands:  

We're having the same issue deploying with Capistrano a newish Rails 5.2 app with Bootsnap 1.4.2, so we temporary disabled Bootsnap. Any plans to cut a new release?

ansonhoyt:
@markets 1.4.3 has been released. I'm re-enabling in production now too.

markets:
Thanks @ansonhoyt will try asap!

Vincent14:
I always encounter the problem after shifting to bootsnap 1.4.5

Issue #168, opened 2018-04-27
IDE debugging of apps on Rails 5.2 broken


grumBit:
### Steps to reproduce
See attached development_machine_installation.txt, then follow steps in expected behaviour section
[development_machine_installation.txt](https://github.com/Shopify/bootsnap/files/1953629/development_machine_installation.txt)

### Expected behavior
In IDE;
- add break point to first line of [app root]/ config/application.rb;
        require_relative 'boot'
- start debug.
- debugger should stop at breakpoint

### Actual behavior
- debugger continues without stopping
- NB: Any breakpoint further into execution of the Rails app fails to halt execution.

### Work around;
Edit [app root]/ config/boot.rb
Comment out line;
- require 'bootsnap/setup'

Debugging now works throughout app as expected.

### Notes;
Looking at https://github.com/Shopify/bootsnap/issues/93, I believe this problem occurs due a combination of the following;
- new Ruby 2.5 feature "Remove `trace` instructions" (https://bugs.ruby-lang.org/issues/14104)
- Rails 5.2 inclusion of bootsnap gem

NB: The same problem is occurring in a number of IDE's;
- Jetbrains;  https://youtrack.jetbrains.com/issue/RUBY-20684
- RubyMine; https://github.com/ruby-debug/ruby-debug-ide/issues/139
- VSCode; https://github.com/rubyide/vscode-ruby/issues/337

I believe this is a Ruby + bootsnap issue rather than bootsnap itself.  However, I'm guessing the "best" place to fix it would be in bootsnap.

Also, I wanted to log the problem with a clear title so other Rails developers using IDE's will know what is going on, rather than spending time on searching for answers

(Apologies if my shotgun approach to submitting this issue on multiple repos is poor-form.  I'm new to open-source development)

### System configuration
- `vscode-ruby` version: 0.18.0
- Ruby version: ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin17]
- Rails version: 5.2.0
- ruby-debug-ide version: 0.6.1
- debase version: 0.2.2
- VS Code version: 1.22.2 (1.22.2)
- Operating System: macOS 10.13.4 (17E199)
- Hardware (optional): MacBook Pro (13-inch, Mid 2012) & MacBook (13-inch, Mid 2010)

tzmfreedom:
Bootsnap use `RubyVM::InstructionSequence.load_from_binary` to load other file quickly, and almost all debugger use `TracePoint` feature.

[...]

PR #257, opened 2019-04-22
Disable iseq cache in Ruby 2.5

rafaelfranca:
There are many bugs in Ruby 2.5 related with iseq cache making it hard to use tracepoints with it. This means that debuggers and any gem that uses tracepoint don't play well with bootsnap when used with Ruby 2.5.

Until those issues are fixed in Ruby 2.5 we are going to disabled iseq cache in the `bootsnap/setup` file.

Fixes https://github.com/rails/rails/issues/35475.

### DIFF
--- a/lib/bootsnap/setup.rb
+++ b/lib/bootsnap/setup.rb
@@ -24,12 +24,15 @@ unless cache_dir
   cache_dir = File.join(app_root, 'tmp', 'cache')
 end
 
+ruby_version = Gem::Version.new(RUBY_VERSION)
+iseq_cache_enabled = ruby_version < Gem::Version.new('2.5.0') || ruby_version >= Gem::Version.new('2.6.0')
+
 Bootsnap.setup(
   cache_dir:            cache_dir,
   development_mode:     development_mode,
   load_path_cache:      true,
   autoload_paths_cache: true, # assume rails. open to PRs to impl. detection
   disable_trace:        false,
-  compile_cache_iseq:   true,
+  compile_cache_iseq:   iseq_cache_enabled,
   compile_cache_yaml:   true,
 )


kaspth:
Won’t this turn on iseq cache for 2.5 only? As far as I understand the logic currently implemented.

casperisfine:
@kaspth good catch. It should be reversed.

rafaelfranca:
Fixed

lib/bootsnap/setup.rb
```diff
@@ -24,12 +24,15 @@
   cache_dir = File.join(app_root, 'tmp', 'cache')
 end
 
+ruby_version = Gem::Version.new(RUBY_VERSION)
+iseq_cache_enabled = ruby_version < Gem::Version.new('2.5.0') || ruby_version >= Gem::Version.new('2.6.0')
```
kaspth: Just for my own curiosity, why are these matchers more correct versus something like `RUBY_VERSION !~ /\A2\.5/`?

rafaelfranca: In this case they are not, I was just following the same pattern other parts of the codebase was using.

kaspth: Ah, I see. All good 👍

burke:
Really the issue is when we're using 2.5.x and also Rails 6+, right? Is there some simple way you can think of to cheaply check that without throwing away performance for apps that haven't updated either?

rafaelfranca:
> Really the issue is when we're using 2.5.x and also Rails 6+, right?

No. Even without Rails 6, if you have iseq cache enable, debugging doesn't work well due to a bug where trace points are not fired when iseq cache is enabled. See https://github.com/Shopify/bootsnap/issues/168 and related issues.

rafaelfranca:
I just saw that they marked @XrXr's to backport to 2.5, that should fix our issue so I'll remove this code when Ruby 2.5.5 is released.

PR #261, opened 2019-06-14
Add helper_test with validating the cache path

jmkim:
This patch will add a test to validate the cache path, by comparing the actual cache path with the path generated by `Help.cache_path()` in `test_helper.rb`.

This patch assumes the `Help.cache_path()` equal with the path generated by C extension.

### DIFF
--- a/test/helper_test.rb
+++ b/test/helper_test.rb
@@ -1,0 +1,12 @@
+require('test_helper')
+
+class HelperTest < MiniTest::Test
+  include(TmpdirHelper)
+
+  def test_validate_cache_path
+    path = Help.set_file('a.rb', 'a = a = 3', 100)
+    cp = Help.cache_path(@tmp_dir, path)
+    load(path)
+    assert_equal(true, File.file?(cp))
+  end
+end


jmkim:
FYI, I already signed CLA https://cla.shopify.com/individual-cla

rafaelfranca:
Thank you for the pull request. Can you explain why this test is useful given it is testing a test helper that will not be used by the users of the library? Which kind of problem is this test trying to catch? Which problems is this test going to catch that the tests on https://github.com/Shopify/bootsnap/blob/61f0639c49d22b3b1868d23f90bc9b125bfbaa89/test/compile_cache_test.rb would not?

jmkim:
Hi, I made this for identifying explicitly if cache path mismatch comes.

I got an error in some test units: `CompileCacheTest#test_invalid_cache_file` and `CompileCacheTest#test_no_write_permission_to_cache`. #263

Now I identified the error; the generated cache path is mismatching between `Help.get_path()` and bootsnap. This patch will show if mismatch problem more explicitly.

And I created new file `helper_test.rb`, as this test unit is verifying `Help.get_path()` in `test_helper.rb` is correct or not.

rafaelfranca:
Thank you for the explanation. It makes sense.

PR #267, opened 2019-07-11
Add support for CRuby 2.7 or later

kou:
CRuby 2.7 or later use Git commit ID as revision. It's String.


kou:
"ruby-head" job failed without this change: https://travis-ci.org/Shopify/bootsnap/jobs/556900107

kou:
"ruby-head" job passes with this change: https://travis-ci.org/Shopify/bootsnap/jobs/557176531

Issue #119, opened 2018-01-13
 Bootsnap::LoadPathCache::Store::NestedTransactionError


jimchild49:
I added bootsnap gem recently to one of my local app, and while the local server was running I went and watched some videos, when I came back to the app and clicked a link I've got  an error (I did no thing everything was running correctly)

Bootsnap::LoadPathCache::Store::NestedTransactionError

it says the error is coming from here : 
```
      def transaction
        raise NestedTransactionError if @in_txn
        @in_txn = true
        yield
      ensure
```

KenkenIchimura:
me too!

rafaelfranca:
Can you please provide a sample application that reproduces the error?

KenkenIchimura:
**Versions** 
bootsnap1.3.2
ruby 2.5.1
Rails 5.2.1

**server/config/boot.rb**

```
ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)

require 'bundler/setup' # Set up gems listed in the Gemfile.
require 'bootsnap/setup' # Speed up boot time by caching expensive operations.

```






**Here is an error message on ReactNative Debugger.**
```
YellowBox.js:67 Possible Unhandled Promise Rejection (id: 10):
TypeError: Network request failed
TypeError: Network request failed
    at XMLHttpRequest.xhr.onerror (blob:file:///be1668be-e3b0-4547-8aee-f37374691cfb:27971:18)
    at XMLHttpRequest.dispatchEvent (blob:file:///be1668be-e3b0-4547-8aee-f37374691cfb:29256:27)
    at XMLHttpRequest.setReadyState (blob:file:///be1668be-e3b0-4547-8aee-f37374691cfb:29009:20)
    at XMLHttpRequest.__didCompleteResponse (blob:file:///be1668be-e3b0-4547-8aee-f37374691cfb:28836:16)
    at blob:file:///be1668be-e3b0-4547-8aee-f37374691cfb:28946:47
    at RCTDeviceEventEmitter.emit (blob:file:///be1668be-e3b0-4547-8aee-f37374691cfb:4538:37)
    at MessageQueue.__callFunction (blob:file:///be1668be-e3b0-4547-8aee-f37374691cfb:2608:44)
    at blob:file:///be1668be-e3b0-4547-8aee-f37374691cfb:2385:17
    at MessageQueue.__guard (blob:file:///be1668be-e3b0-4547-8aee-f37374691cfb:2561:13)
[...]

PR #270, opened 2019-08-15
Update `Bootsnap::LoadPathCache::Store` to better handle concurrent transactions

nalessi:
In our build system we occasionally see the error
```
Bootsnap::LoadPathCache::Store::NestedTransactionError:
  Bootsnap::LoadPathCache::Store::NestedTransactionError
...
```

However, there is no other call to `Bootsnap::LoadPathCache::Store#transaction` in the callstack. This suggests that we have multiple threads modifying the store at the same time.

Reading `Bootsnap::LoadPathCache::Cache` concurrency is already being accounted for with a `Mutex`. Which means this isn't concurrent calls to the same `Bootsnap::LoadPathCache::Cache`. And looking at `Bootsnap::LoadPathCache.setup` there is a single `Bootsnap::LoadPathCache::Store` that is used by `@load_path_cache` and, if active support in `@autoload_paths_cache`.

Given this `Bootsnap::LoadPathCache::Store` needs to be able to deal with multithreading. This replaces `@in_txn` with a mutex which gives the desired synchronization.

This should not lead to deadlocks with the `Bootsnap::LoadPathCache::Cache` mutexes because the `Bootsnap::LoadPathCache::Store` doesn't make any calls into the `Cache` and the `Cache` already makes all `Bootsnap::LoadPathCache::Store#transaction` calls in a `Mutex#synchronize` block.

This may be related to https://github.com/Shopify/bootsnap/issues/119 since the example stack there also only has `transaction` in it once.

### DIFF
--- a/lib/bootsnap/load_path_cache/store.rb
+++ b/lib/bootsnap/load_path_cache/store.rb
@@ -11,7 +11,8 @@ module Bootsnap
 
       def initialize(store_path)
         @store_path = store_path
-        @in_txn = false
+        # TODO: Remove conditional once Ruby 2.2 support is dropped.
+        @txn_mutex =  defined?(::Mutex) ? ::Mutex.new : ::Thread::Mutex.new
         @dirty = false
         load_data
       end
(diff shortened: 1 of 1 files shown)


burke:
We just need you to click a few buttons for that CLA check unfortunately.

nalessi:
@burke Thanks for the review! I've fixed the CLA so this should be ready for a final review.

burke:
Excellent, thanks!

PR #271, opened 2019-08-29
Only strip elidable extensions to make short forms

jeffparsons:
Otherwise we'll erroneously think that 'descriptor' ('descriptor.rb')
is a loaded feature when somebody requires 'descriptor.pb' ('descriptor.pb.rb'),
when in fact those are two completely unrelated files.

### DIFF
--- a/lib/bootsnap/load_path_cache/loaded_features_index.rb
+++ b/lib/bootsnap/load_path_cache/loaded_features_index.rb
@@ -40,7 +40,7 @@ module Bootsnap
             # /a/b/lib/my/foo.rb
             #          ^^^^^^^^^
             short = feat[(lpe.length + 1)..-1]
-            stripped = strip_extension(short)
+            stripped = strip_extension_if_elidable(short)
             @lfi[short] = hash
             @lfi[stripped] = hash
           end
@@ -94,13 +94,14 @@ module Bootsnap
 
         hash = long.hash
 
-        # do we have 'bundler' or 'bundler.rb'?
-        altname = if File.extname(short) != ''
-          # strip the path from 'bundler.rb' -> 'bundler'
-          strip_extension(short)
+        # Do we have a filename with an elidable extension, e.g.,
+        # 'bundler.rb', or 'libgit2.so'?
+        altname = if is_extension_elidable(short)
+          # Strip the extension off, e.g. 'bundler.rb' -> 'bundler'.
+          strip_extension_if_elidable(short)
         elsif long && (ext = File.extname(long))
-          # get the extension from the expanded path if given
-          # 'bundler' + '.rb'
+          # We already know the extension of the actual file this
+          # resolves to, so put that back on.
           short + ext
         end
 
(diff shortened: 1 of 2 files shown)


jeffparsons:
@burke A little bit of background in case you're curious: we have a large number of generated protobuf classes in files named `whatever.pb.rb`, and some of those names are common enough that there will also be an unrelated `whatever.rb` under the same require-root-relative path.

The bug showed up as bootsnap thinking that `whatever.rb` had already been loaded and so doing nothing when `require 'whatever'` was run.

(I've signed the CLA; the bot just hasn't picked it up yet.)

jeffparsons:
Any concerns about this fix? I'm happy to make followup changes if I've missed something...

(CC @rafaelfranca in case @burke isn't around...)

jeffparsons:
(Closed and re-opened to try to trick CLA bot into noticing that I signed the CLA...)

jeffparsons:
@burke No worries. I've just returned from vacation in turn! 🏝 

Oops — I completely forgot about shared libraries. (I've somehow managed to avoid ever requiring them directly.) I've fixed this PR to handle them.

I've made the assumption here that people won't refer to their Ruby files by extremely weird names, e.g., "libgit2.dylib.rb" which is then required as "libgit2.dylib". I'd be willing to bet that such adversarial examples don't actually exist in the wild, but strictly speaking this doesn't match the behaviour of Ruby itself in all cases.

Ruby appears to have quite a complex scheme for resolving require paths (https://ruby-doc.org/core-2.6.4/Kernel.html#method-i-require), including letting you refer to shared libraries "colloquially" using the default shared library extension from whatever platform you want, e.g. 'bruce.dll' on MacOS or 'bruce.so' on Windows. And it's not super-clear to me what order different extensions will be tried on each platform, so it may be quite difficult to match Ruby's behaviour precisely here.

LMK if you're concerned about this, and I can have a think about ways to tackle it. If you think it's important, then in the worst case Bootsnap could carefully track the exact implementation of Ruby release-by-release and replicate the exact behaviour of whatever platform and version of Ruby you're currently running. 🤷‍♂

lib/bootsnap/load_path_cache/loaded_features_index.rb
```diff
@@ -117,8 +118,30 @@ def register(short, long = nil)
...
+      # So optimistically pretend that all known elidable extensions
+      # will be tried on all platforms, and that people are unlikely
+      # to name files in a way that assumes otherwise.
+      # (E.g. It's unlikely that someone will know that their code
+      # will _never_ run on MacOS, and therefore think they can get away
+      # with callling a Ruby file 'x.dylib.rb' and then requiring it as 'x.dylib'.)
+      #
+      # See <https://ruby-doc.org/core-2.6.4/Kernel.html#method-i-require>.
+      def is_extension_elidable(f)
+        f.end_with?('.rb', '.so', '.o', '.dll', '.dylib')
```
burke: My hunch is that this would be faster as `File.extname(f) =~ /^\.(rb|so|o|dll|dylib)$/`

jeffparsons: My (extremely unscientific 😊) microbenchmark of doing 10 million comparisons against a known filename suggests that using `end_with?` is faster. (About 26% of how long the regex match takes.)

This matches my intuition: in the happy case (`*.rb`), Ruby should only need to compare three characters, starting from a known offset. That's pretty hard to beat.

Out of curiosity I tried switching the extensions around in the regex version, and it doesn't seem to make a difference. So I'd say not being able to hint that the vast majority of files will be `*.rb` is hurting pretty badly there.

I also tried using `File.extname` vs. `end_with?` on the full filename, and using `File.extname` is significantly slower. I assume what's happening here is that File.extname ends up allocating memory for the new string, whereas `end_with?` can just compare it to a constant in-place.

burke:
Hmm, ok, I think this looks good then. Thanks for the contribution!

PR #274, opened 2019-09-20
Remove cast from const pointer to non-const pointer

Ryooooooga:


Ryooooooga:
I signed the CLA

burke:
Thanks, this is definitely an improvement.

Issue #277, opened 2019-10-02
LoadPathCache scan takes a lot of time


katafrakt:
I noticed that 2.5 out of 10 seconds of my application's startup time after requiring bootsnap is consumed by `Bootsnap::LoadPathCache::PathScanner.call`. By tracking the issue down, I found that it checks more than 150000 files that `Dir.glob(path + ALL_FILES)`. Most of those files are `node_modules`, cache files and compiled assets.

This is obviously unnecessary. By simply hardcoding skipping of those path I gained 20-25% less of application boot time. I was wondering how this can be parameterized without actually overwriting part of Bootsnap code. 

"All-in" approach would be writing a C extension that would only return those paths that are not explicitly excluded from load path optimization instead of `Dir.glob`. I have some preliminary (working) result here: https://github.com/Shopify/bootsnap/compare/master...katafrakt:dir-walker?expand=1

Lighter approach would be to just `next if excluded_paths.any? {|ep| absolute_path.start_with?(ep)}` to the block inside `Dir.glob.each`, but of course this would be a bit less performant (I don't know how much less).

What are your thoughts about this issue?

burke:
I think something like this makes sense in general, and I'm even okay with a first version of it hardcoding `node_modules` as the only blacklisted item, but I don't have a strong intuition of what the right implementation strategy is.

katafrakt:
I did some more profiling on different solutions:
* just excluding node_modules (`next if relative_path.starts_with?('node_modules')` gives about 85% of time improvement (nice!)
* having a configurable list of absolute paths (5 elements) gives maybe 5-10%
* same with relative paths (so, shorter ones) gives me about 35%
* C extension instead of Dir.glob seems to give 98.5% of time improvement

mscrivo:
Being able to exclude folders, particularly node_modules, would be super nice. I'm playing around with swapping yarn with pnpm in our monorepo and because of pnpm's nested symlinking structure it uses, it makes bootsnap absolutely choke. Startup time went from 4s to approx 1m with pnpm's structure.

katafrakt:
Probably extracting just a config option to exclude some directories from the linked PR would be pretty easy.

PR #278, opened 2019-10-10
Speed up LoadPathCache with more C

katafrakt:
I decided to keep up the momentum with #277 and create a tenative pull request with my proposed changes to load path cache.

This includes:
- Additional option to exclude some path to try them as potential load path to scan
- New module in C to replace `Dir.glob`. It is both faster and easier on memory - unlike `Dir.glob` creating a large array and then iterating it, it yields with every found path and does path exclusion on C level.
- Extending TravisCI matrix to test both with this new extension and without

I tested those changes against Discourse, but without significant speedup:

|                     | cold | warm |
|---------------------|------|------|
| no exclusions, Ruby | 3.49 | 3.19 |
| no exclusions, C    | 3.48 | 3.2  |
| exclusions, Ruby    | 3.62 | 3.13 |
| exclusions, C       | 3.48 | 3.15 |

However, with my internal project with **a lot** of javascript (webpacker etc.), caching etc. savings are significant:

|                     | cold |       | warm |       |
|---------------------|------|-------|------|-------|
| no exclusions, Ruby | 9.45 | 0%    | 8.5  | 0%    |
| no exclusions, C    | 9.02 | 4.5%  | 7.94 | 6.5%  |
| exclusions, Ruby    | 8.49 | 10.1% | 6.83 | 19.6% |
| exclusions, C       | 6.92 | 26.7% | 6.86 | 19.2% |

**Legend:**
- cold - with load path cache purged (`rm tmp/cache/bootsnap-load-path-cache`)
- warm - with load path cache present and no changes to the files

I'm not the best C programmer (or even mediocre one), so to use this new C extension one must set environment variable `BOOTSNAP_EXPERIMENTAL`. This way people with project in dire need to speed up can experimentally use this, while most of the people can safely use existing version.

Let me know what you think about this direction.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ rvm:
   - ruby-2.4
   - ruby-2.5
   - ruby-head
+env:
+  - ENABLE_EXPERIMENTAL=1
+  - ''
 
 matrix:
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ rvm:
   - ruby-2.5
   - ruby-head
 env:
-  - ENABLE_EXPERIMENTAL=1
+  - BOOTSNAP_EXPERIMENTAL=1
   - ''
 
 matrix:
--- a/Rakefile
+++ b/Rakefile
@@ -9,5 +9,11 @@ Rake::ExtensionTask.new do |ext|
   ext.lib_dir = 'lib/bootsnap'
   ext.gem_spec = gemspec
 end
+Rake::ExtensionTask.new do |ext|
+  ext.name = 'dirscanner'
+  ext.ext_dir = 'ext/dirscanner'
+  ext.lib_dir = 'lib/bootsnap'
+  ext.gem_spec = gemspec
+end
 
 task(default: :compile)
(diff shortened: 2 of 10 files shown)


katafrakt:
Rebased master to resolve conflict

h0jeZvgoxFepBQ2C:
Is there any ETA when this gets merged? We woul profit from this probably ❤️

katafrakt:
@h0jeZvgoxFepBQ2C this now conflicts with changes from #312. I think even with the other PR merged it would be still beneficial to have an option to exclude some dirs (node_modules, cache). I can extract it from this one and create another pull request if the maintainers are interested.

Pure C dir walker of course still would be faster than current master version, but the combination of changes in #312 with exclusion option might be enough to give significant speedup without the necessity to maintain scary chunk of C code.

casperisfine:
The important part, which is skipping big useless directories like `node_modules` was implemented in https://github.com/Shopify/bootsnap/pull/424.

As for directory scanning in C, I doesn't seem like it yields enough of a speedup to justify the increased complexity. Additionally the branch was very udated so the rebasing would have been tricky.

Issue #75, opened 2017-07-27
Cannot define multiple 'included' blocks for a Concern


Aqualon:
After seeing https://github.com/rails/rails/pull/29313 I wanted to give bootsnap a try, created a new rails 5.1.2 app on ruby 2.4.1 via ```rails new test_bootsnap```. I added bootsnap to Gemfile and boot.rb like in the pull request and tried to start rails console via `bundle exec rails c` resulting in the following error:

```
$ bundle exec rails c
/Users/aqualon/.rvm/gems/ruby-2.4.1/gems/activesupport-5.1.2/lib/active_support/concern.rb:126:in `included': Cannot define multiple 'included' blocks for a Concern (ActiveSupport::Concern::MultipleIncludedBlocks)
	from /Users/aqualon/.rvm/rubies/ruby-2.4.1/lib/ruby/gems/2.4.0/gems/actionview-5.1.2/lib/action_view/view_paths.rb:5:in `<module:ViewPaths>'
	from /Users/aqualon/.rvm/rubies/ruby-2.4.1/lib/ruby/gems/2.4.0/gems/actionview-5.1.2/lib/action_view/view_paths.rb:2:in `<module:ActionView>'
	from /Users/aqualon/.rvm/rubies/ruby-2.4.1/lib/ruby/gems/2.4.0/gems/actionview-5.1.2/lib/action_view/view_paths.rb:1:in `<main>'
	from /Users/aqualon/.rvm/rubies/ruby-2.4.1/lib/ruby/gems/2.4.0/gems/bootsnap-1.1.2/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
	from /Users/aqualon/.rvm/rubies/ruby-2.4.1/lib/ruby/gems/2.4.0/gems/bootsnap-1.1.2/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
	from /Users/aqualon/.rvm/gems/ruby-2.4.1/gems/activesupport-5.1.2/lib/active_support/dependencies.rb:292:in `block in require'
	from /Users/aqualon/.rvm/gems/ruby-2.4.1/gems/activesupport-5.1.2/lib/active_support/dependencies.rb:258:in `load_dependency'
	from /Users/aqualon/.rvm/gems/ruby-2.4.1/gems/activesupport-5.1.2/lib/active_support/dependencies.rb:292:in `require'
[...]

PR #283, opened 2019-12-07
Use glibc version in ruby_platform cache header if available

stanhu:
Previously, the `ruby_platform` header in the cache was calculated as a
hash of the `RUBY_PLATFORM` (e.g. x86_64-linux-gnu) with the `uname -v`
output (e.g. `#75-Ubuntu SMP Tue Oct 1 05:24:09 UTC 2019`).

However, using `uname -v` is a bit restrictive on Linux systems. It
makes it difficult for a containerized application to bundle the
Bootsnap cache with the application itself. For example, in Docker, the
host operating system determines the kernel version, so even though a
Docker image may be built with a specific base image, the output of
`uname -v` will change depending on where it's run. As a result, the
application will need to rebuild the Bootsnap cache every time unless a
persistent volume is used AND the kernel version is fixed. However, both
requirements can be difficult to guarantee, which negates any speed-up
that Bootsnap provides.

We observe that in a Linux system, the application binary interface
(ABI) is defined by the kernel version and glibc version. A change in
the kernel ABI requires a change to glibc. Instead of hashing the output
of `uname -v`, we hash the glibc version if it is defined by the
compiler.

### DIFF
--- a/README.jp.md
+++ b/README.jp.md
@@ -152,7 +152,7 @@ close     n
 Bootsnap は、64バイトのヘッダーとそれに続くキャッシュの内容を含んだキャッシュファイルを書き込みます。ヘッダーは、次のいくつかのフィールドで構成されるキャッシュキーです。
 
 - `version`、Bootsnapにハードコードされる基本的なスキーマのバージョン
-- `ruby_platform`、`RUBY_PLATFORM`（x86_64-linux-gnuなど）変数とOSバージョンのハッシュ（Linux、BSD、macOSでは` uname -v`）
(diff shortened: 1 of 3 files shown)


WarheadsSE:
:wave: We'd much appreciate this change, as this has a huge impact on containerized applications with a single execution during the lifetime of the container instance.

WarheadsSE:
@burke Is there anything you'd like in terms of clarification?

burke:
LGTM, sorry for the delay.

stanhu:
@rafaelfranca @burke Thanks! Would you mind releasing a new version of Bootsnap?

WarheadsSE:
For context on @stanhu 's request. We've implemented a container including the changes in order to test real-world implications (https://gitlab.com/gitlab-org/build/CNG/-/merge_requests/376). We've found it quite viable, and would prefer to include a released & version locked Gem.

PR #287, opened 2019-12-25
Add runtime dependency to thwait and e2mmap gems

zunda:
These gem are no longer bundled to Ruby as of ruby-2.7.0
https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/

### DIFF
--- a/bootsnap.gemspec
+++ b/bootsnap.gemspec
@@ -43,4 +43,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency("mocha", "~> 1.2")
 
   spec.add_runtime_dependency("msgpack", "~> 1.0")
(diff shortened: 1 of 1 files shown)


zunda:
I signed the CLA :)

rafaelfranca:
Where are we using those dependencies in the codebase? I could not find.

zunda:
Thanks for checking and sorry for the noise! This turned out to be a red herring.

Because `rake -P` failed with `LoadError` at `lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:34` in `require`, I though bootsnap tried to require the uninstalled library. It should actually be another gem that is requiring it through bootsnap.

PR #289, opened 2020-01-06
allow msgpack 1.1.X, 1.2.X and 1.3.X

bastelfreak:


bastelfreak:
oh I think I got the version requirements wrong. `~> 1.0` already allows every `1.x` version, so this PR doesn't add any value?

rafaelfranca:
That is right. ~> 1.0 already support all 1.x versions. Thank you for the pull request

PR #296, opened 2020-04-06
Add TruffleRuby back to CI

kipply:
It passes now! This commit is a revert of https://github.com/Shopify/bootsnap/commit/81dc814fc9967852d29d30150bb1eb940b8c75e2

cc @eregon


eregon:
Thanks!

Issue #177, opened 2018-07-12
Errno::ENOENT - No such file or directory - bs_fetch:atomic_write_cache_file:rename


Ferdy89:
Hi! We've recently started to use Bootsnap (1.3.0) and we just noticed this error pop up in our Sentry account for our staging server:

```
Errno::ENOENT
No such file or directory - bs_fetch:atomic_write_cache_file:rename

bootsnap/compile_cache/iseq.rb in fetch at line 37
bootsnap/compile_cache/iseq.rb in load_iseq at line 37
bootsnap/load_path_cache/core_ext/kernel_require.rb in require at line 21
bootsnap/load_path_cache/core_ext/kernel_require.rb in block in require_with_bootsnap_lfi at line 21
bootsnap/load_path_cache/loaded_features_index.rb in register at line 65
bootsnap/load_path_cache/core_ext/kernel_require.rb in require_with_bootsnap_lfi at line 20
bootsnap/load_path_cache/core_ext/kernel_require.rb in require at line 29
```

Both the caller and the required file are from gems. We run Rails on a single-threaded/single-process Linux environment inside Docker. So far it has only happened once.

I don't have enough knowledge about Bootsnap to dive deeper into the issue, but I'm happy to provide any other information I can. However, we haven't been able to reproduce it.

Thank you!

burke:
(sorry for the extreme delay in response)

I'm looking through the code and there are two reasons I can think of for this.

Basically what's happening is:

1. We want to write out `/a/b/c`, so we write out `/a/b/c.tmp.12384`, creating `/a/b` if necessary.
2. We move `/a/b/c.tmp.12384` to `/a/b/c`.

The error (`ENOENT`) is coming from step 2.

This means one of two things:

1. `/a/b/c.tmp.12384` didn't exist; or
2. `/a/b` didn't exist.

The two reasons I can think of for this happening are:

1. A path over 1000 characters long made us create the temporary path incorrectly; or
2. The tempfile was somehow deleted during the execution of this function.

Anyone who's experiencing this, issue, do you have extremely long paths? Or do you have some kind of frequently-running cruft-cleanup task?

Ferdy89:
Thank you for looking into this! I looked through our logs and we're still seeing this issue sporadically (fortunately, still only in our staging environment!).

Here are some data points from the most recent occurrences:

* It sometimes happens when loading fairly long paths, but not _extremely_ long. Ours is about 127 characters long
[...]

Issue #174, opened 2018-06-11
TOCTOU Vulnerability and mktemp


junaruga:
When I ran a lint tool for bootsnap for Fedora RPM packaging.

I got below issue (vulnerability) about `mktemp` function used in `bootsnap`.

```
$ rpmlint -i ./*.spec /var/lib/mock/fedora-rawhide-x86_64/result/*.rpm
...
rubygem-bootsnap.x86_64: E: call-to-mktemp /usr/lib64/gems/ruby/bootsnap-1.3.0/bootsnap/bootsnap.so
This executable calls mktemp. As advised by the manpage (mktemp(3)), this
function should be avoided. Some implementations are deeply insecure, and 
there is a race condition between the time of check and time of use (TOCTOU).
See http://capec.mitre.org/data/definitions/29.html for details, and contact
upstream to have this issue fixed.
```

It seems that this vulnerability is related to predictable temporary file path. [1][2]
Right now the created temp file path might look hard to predict.

```
$ bundle exec ruby -e "require 'bootsnap'; Bootsnap.setup(cache_dir: '/tmp/foo', autoload_paths_cache: false)"
```
=>
ext/bootsnap/bootsnap.c
atomic_write_cache_file tmp_path is /tmp/foo/bootsnap-compile-cache/c9/1fac3dd22d6c1c.tmp.5EfzIQ

But have you considered about the vulnerability?
How do you think about the vulnerability?

I tried to implement with `mkstemp` instead of `mktemp`.
Though I have no idea about how to access the directory later in the case like below code.

```
    Bootsnap::CompileCache::Native.fetch(@tmp_dir, file, TestHandler)
    data = File.read(Help.cache_path(@tmp_dir, file))
```

https://github.com/junaruga/bootsnap/commit/0dbd8c3a354fa0ad88178d472dcbe4b247ca9e17

[1] https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use
[2] https://www.owasp.org/index.php/Insecure_Temporary_File

burke:
Sorry for the extreme delay in implementing this. [I've opened a PR to address this](https://github.com/Shopify/bootsnap/pull/240); I'm going to close this issue just to get the backlog cleaned up, happy to discuss over there.

PR #309, opened 2020-06-28
Ensure that temporary file is used only by one process

abicky:
`mkstemp(3)` ensures that a unique file is created, but in the previous implementation, there's a possibility that a process
uses the temporary file created by another process if `mkstemp(3)` fails to create a file due to EEXIST. That has the same risk as https://github.com/Shopify/bootsnap/issues/174.

This PR will also resolve https://github.com/Shopify/bootsnap/issues/177 if the cause is that multiple processes try to create a file with the same name at the same time.

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -32,6 +32,8 @@
 
 #define KEY_SIZE 64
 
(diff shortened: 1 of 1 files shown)


matti:
How about also releasing the gem, the latest release is 5 months old?

PR #314, opened 2020-07-31
Prevent FallbackScan from polluting exception cause

aaronjensen:
When debugging an issue with [test-bench](https://github.com/test-bench/test-bench) and bootsnap, I discovered that bootsnap pollutes error.cause by requiring the file naturally after a fallback scan within a rescue. This causes any error that is thrown within the required file to have a `#cause` defined, which it should not. This improves output in test-bench when an error occurs and it may (I haven't tested this) improve output in other frameworks that inspect error.cause.

I wasn't sure how to write a test for this because I didn't know how to trigger a fallback scan (I don't even really know what that is) and still load a real file that raises an exception. If you could give me a snippet for that I can add a test.

Thanks!

### DIFF
--- a/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
+++ b/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
@@ -38,7 +38,11 @@ module Kernel
   rescue Bootsnap::LoadPathCache::ReturnFalse
     false
   rescue Bootsnap::LoadPathCache::FallbackScan
-    require_with_bootsnap_lfi(path)
+    fallback = true
+  ensure
+    if fallback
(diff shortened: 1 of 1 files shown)


sbellware:
The maintainer might just take this code as a proof-of-concept and make the changes themselves. That would obviate the CLA hoop-jumping and delays.

burke:
LGTM. We'll live without a test for this.

sbellware:
Awesome ❤️

aaronjensen:
Thank you for the merge!

If you have a free moment to cut a new gem release, I'd really appreciate it. Thank you.

burke:
Done!

PR #320, opened 2020-09-10
CompileCache::YAML add support for symbolize_names and freeze options

casperisfine:
So far the YAML cache has been simple enough to implement because `YAML.load_file` pretty much only take a single parameter. However it's mostly an oversight. Psych `3.1` added a `symbolize_names:` options to `load`, and `3.2` a `freeze:` option.

In https://github.com/ruby/psych/pull/463 I'm requesting these options to be forwarded.

For `symbolize_keys` it's easy since message pack already has the same option since a very long time, and anyway it supports symbols so that will work transparently.

`freeze` is a bit more tricky, as `MessagePack` need to do the same thing. I have a PR for that https://github.com/msgpack/msgpack-ruby/pull/194 and the maintainer agreed on it, so it should land at some point. So with a bit of feature testing we're fine.

On another note, it's not yet released, but [`json` added a `load_file` method as well](https://github.com/flori/json/pull/387) which I'd like to add supports for, and it takes options too. So this change is kind of a two birds one stone.

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -91,16 +91,16 @@ static ID uncompilable;
 
 /* Functions exposed as module functions on Bootsnap::CompileCache::Native */
 static VALUE bs_compile_option_crc32_set(VALUE self, VALUE crc32_v);
-static VALUE bs_rb_fetch(VALUE self, VALUE cachedir_v, VALUE path_v, VALUE handler);
+static VALUE bs_rb_fetch(VALUE self, VALUE cachedir_v, VALUE path_v, VALUE handler, VALUE args);
 
 /* Helpers */
 static uint64_t fnv1a_64(const char *str);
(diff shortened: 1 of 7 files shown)


tenderlove (approved):
This makes sense to me (including the xor on the hash).  One thing though, rather than passing the string buffer in to `bs_cache_path` should we just pass the hash value?  That way we can do something more complex in case we don't have just strings.

Also, should we use `rb_str_hash` on the string instead of getting the buffer and calling `fnv1a_64`?

Note that I think this is fine as-is. 😀

casperisfine:
> Also, should we use `rb_str_hash` on the string instead of getting the buffer and calling `fnv1a_64`?

I figured the same thing, but then I realized that `#hash` is randomly seeded when Ruby boots, so it wouldn't stay consistent:

```
$ ruby -e 'p "foo".hash'
-4056315210992301503
$ ruby -e 'p "foo".hash'
4213994398207020531
```

Thanks for the review!

casperisfine:
I just tested it against our main app, works as expected.

PR #326, opened 2020-10-28
Add a bootsnap command to allow to precompile gems without booting the application

casperisfine:
In normal usage, bootsnap's iseq cache is computed on the fly as the ruby files are required.

Which means that if you want to precompute the cache, the only way is to boot the application.

In my case I'm building Docker image for ruby applications, and I'm caching all the gems in a content addressed layer. So I would like to be able to precompile the gems iseq cache without having to have a bootable application, so that the bootsnap cache is part of the bundler layer.

This is exactly what this new executable allows, it setup bundler to get all the relevant gems in `$LOAD_PATH`, and then precompile all the requirable.

### DIFF
--- a/lib/bootsnap/cli.rb
+++ b/lib/bootsnap/cli.rb
@@ -1,0 +1,136 @@
+# frozen_string_literal: true
+
+require 'bootsnap'
+require 'optparse'
+require 'fileutils'
+
+module Bootsnap
+  class CLI
+    unless Regexp.method_defined?(:match?)
+      module RegexpMatchBackport
+        refine Regepx do
+          def match?(string)
+            !!match(string)
+          end
+        end
+      end
+      using RegexpMatchBackport
+    end
+
+    attr_reader :cache_dir, :argv
+
+    attr_accessor :compile_gemfile, :exclude
+
+    def initialize(argv)
+      @argv = argv
+      self.cache_dir = 'tmp/cache'
+      self.compile_gemfile = false
+      self.exclude = nil
+    end
+
+    def precompile_command(*sources)
+      require 'bootsnap/compile_cache/iseq'
+
+      Bootsnap::CompileCache::ISeq.cache_dir = self.cache_dir
+
+      if compile_gemfile
+        sources += $LOAD_PATH
+      end
+
+      sources.map { |d| File.expand_path(d) }.each do |path|
+        if !exclude || !exclude.match?(path)
+          list_ruby_files(path).each do |ruby_file|
+            if !exclude || !exclude.match?(ruby_file)
+              CompileCache::ISeq.fetch(ruby_file, cache_dir: cache_dir)
+            end
+          end
+        end
+      end
+      0
+    end
+
+    dir_sort = begin
+      Dir['.', sort: false]
+      true
+    rescue ArgumentError
+      false
+    end
+
+    if dir_sort
+      def list_ruby_files(path)
+        if File.directory?(path)
+          Dir[File.join(path, '**/*.rb'), sort: false]
+        elsif File.exist?(path)
+          [path]
+        else
+          []
+        end
+      end
+    else
+      def list_ruby_files(path)
+        if File.directory?(path)
+          Dir[File.join(path, '**/*.rb')]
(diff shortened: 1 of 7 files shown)


rafaelfranca (approved):

jmreid (approved):

DazWorrall (approved):
Nice idea 👍

burke (approved):
Great idea :+1:

lib/bootsnap/cli.rb
```diff
@@ -0,0 +1,120 @@
...
+
+require 'bootsnap'
+require 'optparse'
+require 'fileutils'
+
+module Bootsnap
+  class CLI
+    unless Regexp.method_defined?(:match?)
+      module RegexpMatchBackport
+        refine Regepx do
```
etiennebarrie, line 11: Since it's just a CLI and the process is thrown away, it might be enough to define the method directly? Or is the code actually loaded? I thought it was just parsed and the iseq cached.

lib/bootsnap/cli.rb
```diff
@@ -0,0 +1,118 @@
...
+      if compile_gemfile
+        sources += $LOAD_PATH
+      end
+
+      CompileCache::ISeq.compile_option_updated
+      iseq_cache_dir = File.join(cache_dir, 'bootsnap-compile-cache')
+
+      source.map { |d| File.expand_path(d) }.each do |path|
+        Dir[File.join(path, '**/*.rb')].each do |ruby_file|
+          if !exclude || !exclude.match?(ruby_file)
```
etiennebarrie, line 50: Or even just use `=~` which does allocate a MatchData if that's what we're going for.

lib/bootsnap/cli.rb
```diff
@@ -0,0 +1,118 @@
...
+        invalid_usage!("Unknown command: #{command}")
+      end
+    end
+
+    private
+
+    def invalid_usage!(message)
+      STDERR.puts message
+      STDERR.puts
+      STDERR.puts parser
```
etiennebarrie, line 79: Maybe `warn` would be simpler?

lib/bootsnap/cli.rb
```diff
@@ -0,0 +1,118 @@
...
+          STDERR.puts "#{source} is not a directory"
+          return 1
+        end
+      end
+
+      if compile_gemfile
+        sources += $LOAD_PATH
+      end
+
+      CompileCache::ISeq.compile_option_updated
```
etiennebarrie, line 45: Not really sure it's worth changing the API or adding an alias, but this name could be clearer, now that it's called from outside of `CompileCache::ISeq`.

etiennebarrie (approved):

lib/bootsnap/cli.rb
casperisfine: I'd rather avoid doing it. I'll likely add some tests for that class, so it's better if it doesn't monkey patch Regexp.

lib/bootsnap/cli.rb
casperisfine: Yeah. I think I'll refactor this so that I don't need this to be used.

casperisfine:
So after trying this on one of our mid sized apps, I made a few changes.

I added  `--exclude PATTERN` option because I noticed some large gems with lots of never loaded files. e.g. `google-api` or `aws-` gems. They contains tons of autogenerated code that most people never require. Without this option the bootsnap cache grew from `~100MB` to `~280MB`. After excluding the couple of outlier gems it is down to a more reasonable `~180MB`.

I also realized that `rake environment` doesn't eager load the app even if `config.eager_load = true` is set, it needs a distinct `config.rake_eager_load = true` option. Until then I was relying on `rake assets:precompile` to eager load the app and warm the bootsnap cache, but now I think it would be preferable to more explicitly precompile the app code. So now the command has a `--gemfile` flag to automatically compile all the gems, and take a list of paths to be precompiled.

I'll add some tests and some documentation tomorrow and likely will cut a new release.

casperisfine:
Hum, it's been hours and the travis builds still didn't start. We should migrate to GitHub Actions soon.

PR #333, opened 2020-11-07
Update input_to_output method arguments to use dissimilar names

connorshea:
This method using `_` for both arguments was causing Sorbet to fail when generating type information (https://github.com/sorbet/sorbet/issues/3561). This change doesn't modify any functionality or outward API in the gem itself. The argument names are based on the arguments for `input_to_output` in `lib/bootsnap/compile_cache/yaml.rb`.


connorshea:
I'm not sure how to get the CLA job to restart but I have signed it

PR #334, opened 2020-11-27
Use a more manageable cache directory structure

casperisfine:
The current cache paths are:

 - `<DIR>/bootsnap-load-path-cache`
 - `<DIR>/bootsnap-compile-cache`

This structure is annoying when you are working on CI systems where you need to specify the paths that need to be persisted across builds, and similar scenarios.

I propose to change it for:

- `<DIR>/bootsnap/load-path-cache`
- `<DIR>/bootsnap/compile-cache`

Making it easier to manage.

On bit problem though is that this change might break some people's workflow, as it's not uncommon to have various script to delete / persist / etc the cache.

@rafaelfranca what do you think?
cc @Shopify/rails

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* Changed the load path cache directory from `$BOOTSNAP_CACHE_DIR/bootsnap-load-path-cache` to `$BOOTSNAP_CACHE_DIR/bootsnap/load-path-cache` for ease of use (#334)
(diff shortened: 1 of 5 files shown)


casperisfine:
Note, the README says:

> you may need to periodically purge `tmp/cache/bootsnap*`

So assuming most users do use a glob like suggested, the change should have no impact.

rafaelfranca (approved):
If this change is clear in the CHANGELOG I don't see a problem to make it.

Minoru-Noda:
qa2 に merge して一回 deploy してみる。無事 deploy できたら ok とする

PR #340, opened 2021-01-25
CLI: also precompile YAML cache

casperisfine:
On some apps, the YAML cache is also a big part of boot performance. Ideally the `precompile` CLI should precompile YAML caches as well.

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -92,6 +92,7 @@ static ID uncompilable;
 /* Functions exposed as module functions on Bootsnap::CompileCache::Native */
 static VALUE bs_compile_option_crc32_set(VALUE self, VALUE crc32_v);
 static VALUE bs_rb_fetch(VALUE self, VALUE cachedir_v, VALUE path_v, VALUE handler, VALUE args, VALUE args_key);
+static VALUE bs_rb_precompile(VALUE self, VALUE cachedir_v, VALUE path_v, VALUE handler);
 
 /* Helpers */
 static uint64_t fnv1a_64(const char *str);
@@ -99,6 +100,7 @@ static void bs_cache_path(const char * cachedir, const char * path, const char *
 static int bs_read_key(int fd, struct bs_cache_key * key);
 static int cache_key_equal(struct bs_cache_key * k1, struct bs_cache_key * k2);
 static VALUE bs_fetch(char * path, VALUE path_v, char * cache_path, VALUE handler, VALUE args);
+static VALUE bs_precompile(char * path, VALUE path_v, char * cache_path, VALUE handler);
 static int open_current_file(char * path, struct bs_cache_key * key, const char ** errno_provenance);
 static int fetch_cached_data(int fd, ssize_t data_size, VALUE handler, VALUE args, VALUE * output_data, int * exception_tag, const char ** errno_provenance);
 static uint32_t get_ruby_revision(void);
@@ -149,6 +151,7 @@ Init_bootsnap(void)
 
   rb_define_module_function(rb_mBootsnap_CompileCache_Native, "coverage_running?", bs_rb_coverage_running, 0);
   rb_define_module_function(rb_mBootsnap_CompileCache_Native, "fetch", bs_rb_fetch, 5);
+  rb_define_module_function(rb_mBootsnap_CompileCache_Native, "precompile", bs_rb_precompile, 3);
   rb_define_module_function(rb_mBootsnap_CompileCache_Native, "compile_option_crc32=", bs_compile_option_crc32_set, 1);
 
   current_umask = umask(0777);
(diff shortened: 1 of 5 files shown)


lib/bootsnap/cli.rb
```diff
@@ -131,6 +172,10 @@ def parser
           self.cache_dir = dir
         end
 
+        opts.on('--verbose', '-v', help.strip) do
+          self.verbose = true
```
casperisfine, line 176: This is helpful to see what gets precompiled and adjust the `--exclude` patterns.

lib/bootsnap/cli.rb
```diff
@@ -117,6 +153,11 @@ def cache_dir=(dir)
       @cache_dir = File.expand_path(File.join(dir, 'bootsnap/compile-cache'))
     end
 
+    def exclude_pattern(pattern)
+      (@exclude_patterns ||= []) << Regexp.new(pattern)
+      self.exclude = Regexp.union(@exclude_patterns)
```
casperisfine, line 158: This allows several `--exclude` arguments rather than just one. Otherwise it can get complex to join patterns.

ext/bootsnap/bootsnap.c
```diff
@@ -331,6 +334,32 @@ bs_rb_fetch(VALUE self, VALUE cachedir_v, VALUE path_v, VALUE handler, VALUE arg
   return bs_fetch(path, path_v, cache_path, handler, args);
 }
 
+/*
+ * Entrypoint for Bootsnap::CompileCache::Native.precompile.
+ * Similar to fetch, but it only generate the cache if missing
+ * and doesn't return the content.
```
casperisfine, line 340: The `precompile` command is a bit slow on large repos/gemfiles, so I think having a dedicated `.precompile` method that does a bit less work is worth it.

lib/bootsnap/cli.rb
```diff
@@ -19,59 +19,62 @@ def match?(string)
...
-              if !exclude || !exclude.match?(ruby_file)
-                CompileCache::ISeq.fetch(ruby_file, cache_dir: cache_dir)
-              end
-            end
-          end
+        if compile_gemfile
+          # Some gems embed their tests, they're very unlikely to be loaded, so not worth precompiling.
+          gem_exclude = Regexp.union([exclude, %r{/spec/}, %r{/test/}].compact)
+          precompile_ruby_files($LOAD_PATH.map { |d| File.expand_path(d) }, exclude: gem_exclude)
+          precompile_yaml_files($LOAD_PATH.flat_map { |p| p.scan(/.*\/gems\/[^\/]+\//) }.uniq, exclude: gem_exclude)
```
etiennebarrie, line 51: For this one the `%r{}` syntax would be useful.  I think `scan` is a bit weird, I don't think we expect multiple results from a single load path.

```suggestion
          precompile_yaml_files($LOAD_PATH.map {|p| p[%r{.*/gems/[^/]+/}] }.compact.uniq, exclude: gem_exclude)
```

Also `/gems/` is [definitely in the path](https://github.com/rubygems/rubygems/blob/b5a113c040a1820a9f0fdd9a11965e695939e2be/lib/rubygems/installer.rb#L262), but technically one of the gem paths could have a /gems/ before the actual gems directory. Maybe we could go off `Gem.path` instead?

Finally the idea of going at the gem top-level dir instead of in the load path is because sometimes gems store YAML files in dirs that are not in the load path like `config/`? Could we add a test or two for this line somehow?

lib/bootsnap/cli.rb
etiennebarrie: 🙏 definitely great feature. The help isn't properly set though it seems.

lib/bootsnap/cli.rb
```diff
@@ -144,7 +189,17 @@ def parser
         help = <<~EOS
           Path pattern to not precompile. e.g. --exclude 'aws-sdk|google-api'
```
etiennebarrie, line 190: Maybe change the help to use multiple patterns? OTOH this shows it supports regexps…

lib/bootsnap/cli.rb
```diff
@@ -144,7 +189,17 @@ def parser
...
         EOS
-        opts.on('--exclude PATTERN', help) { |pattern| self.exclude = Regexp.new(pattern) }
+        opts.on('--exclude PATTERN', help) { |pattern| exclude_pattern(pattern) }
+
+        help = <<~EOS
+          Disable Iseq (.rb) precompilation.
```
etiennebarrie, line 195:
```suggestion
          Disable ISeq (.rb) precompilation.
```

etiennebarrie (approved):

lib/bootsnap/cli.rb
casperisfine: > I think scan is a bit weird, I don't think we expect multiple results from a single load path.

We indeed don't. But it synergize surprisingly well with `flat_map`, I somehow find it elegant. But maybe a `String.sub` might be simpler. I'll dig that a bit.

> Maybe we could go off Gem.path instead?

Hum, yeah I'll check that. The `/gem/` is indeed a bit too hacky.

> Finally the idea of going at the gem top-level dir instead of in the load path is because sometimes gems store YAML files in dirs that are not in the load path like config/?

Yes, exactly.

> Could we add a test or two for this line somehow?

I'll see about that.

rafaelfranca (approved):

lib/bootsnap/cli.rb
etiennebarrie: yeah I'm not a fan of map/compact either, I would have gone with filter_map but I think we want to keep support for 2.6. But other than that `String#[](Regexp)` is kind of a single element `String#scan(Regexp)`, so it fits well. But it's just nitpicks anyway, whatever works best for you.

PR #346, opened 2021-02-03
Fix windows build

casperisfine:
Windows builds started failing. Turns out it's an old bug in the test suite that only surfaced because some cache keys changed.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -60,17 +60,13 @@ module MiniTest
   class Test
     module Help
       class << self
-        def binary(str)
-          str.force_encoding(Encoding::BINARY)
-        end
-
         def cache_path(dir, file, args_key = nil)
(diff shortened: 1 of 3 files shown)


test/test_helper.rb
```diff
@@ -60,17 +60,13 @@ module MiniTest
...
-        end
-
         def cache_path(dir, file, args_key = nil)
           hash = fnv1a_64(file)
           unless args_key.nil?
             hash ^= fnv1a_64(args_key)
           end
 
-          hex = hash.to_s(16)
+          hex = hash.to_s(16).rjust(16, '0')
```
casperisfine, line 69: This is the main fix. `to_s(16)` If the first byte was 0, the returned string wouldn't be 16 char long, which would mess with the cache key generation.

Issue #355, opened 2021-04-23
Precompilation & Load Path Cache


metaskills:
I am working on optimizing some cold starts for Rails in AWS Lambda (https://lamby.custominktech.com) and wonder if you could describe why the [precompilation command](https://github.com/Shopify/bootsnap#precompilation) does not also add the `load-path-cache`. My goal is to have this in place for the Docker image deployed to Lambda so Rails boots faster. 

I've come up with a workaround to do just that. But hoping y'all can add some context as to how the load path cache is supposed to work.

casperisfine:
I think I didn't add it because that cache is dependent on the `$LOAD_PATH`, and that it's hard to predict an app `$LOAD_PATH` without booting it.

But if you have some ideas or rough patches, feel free to submit them.

metaskills:
Another thing to note is that Lambda is a read-only file system. My method to build the `load-path-cache` is to simply add something like this to the Dockerfile.

```dockerfile
RUN bundle exec bootsnap precompile --gemfile app/ lib/ && \
    LAMBY_BUILD=1 bundle exec ruby app.rb
```

Where `app.rb` here does the typical requires for Rails boot, application, environment and defines the Lambda handler. This does generate a good looking `load-path-cache` within the Lambda Container. Still testing benchmarks. 

However, one big hurdle I had to overcomes is that this file is opened and re-written on boot. I've done some general testing and it appears to not change. Can you share why this file is mutated and ends up `@dirty` during `commit_transaction`? I have to add code like this to the `app.rb` to work around it while I test things further. 

```ruby
unless ENV['LAMBY_BUILD']
  require 'bootsnap'
  Bootsnap::LoadPathCache::Store.class_eval do
    def commit_transaction ; end
  end
end
```

casperisfine:
> to simply add something like this to the Dockerfile.

That's pretty much what we do too. Mostly to run `rake assets:precompile` but it generates the load path cache.

> Can you share why this file is mutated and ends up @dirty during commit_transaction?

Hum, that sounds weird. I'd have to dig that up. Not sure I'll have the time today though.

Does your app ever mutate `$LOAD_PATH` ?

metaskills:
Would a different `$LOAD_PATH` on the Docker container vs the Lambda runtime environment cause this? For example, I looked at both and saw these differences.

```
>> d - l
=> ["/var/lang/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib"]
>> l - d
[...]

PR #358, opened 2021-04-26
Ignore Read-Only Filesystems

metaskills:
In repose to https://github.com/Shopify/bootsnap/issues/355 where using the Load Path Cache with AWS Lambda results in small differences with the `$LOAD_PATH` and tries to dumps new data to a read-only filesystem. We can ignore this error but I thought I would outline a few questions that others might have.

<dl>
  <dt>Can you use Lambda Containers `/tmp` vs the `/var/task/tmp` directory?</dt>
  <dd>No. Building a container with a `/tmp` directory will be automatically emptied out when the container image instance is loaded by Firecracker.</dd>
  <dt>Can you copy on boot to the `/tmp` directory?</dt>
  <dd>Yes. But that would negate most performance wins. </dd>
</dl>

## Example Container/Runtime `$LOAD_PATH` Differences

Assuming `l` is the Lambda Runtime load path and `d` is the ECR image via Docker. The differences would be.

```
>> d - l
=> ["/var/lang/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib"]
>> l - d
=> ["/var/task", "/var/runtime/lib", "/opt/ruby/lib", "/var/runtime/gems/bundler-2.2.16/lib"]
```

### DIFF
--- a/lib/bootsnap/load_path_cache/store.rb
+++ b/lib/bootsnap/load_path_cache/store.rb
@@ -91,7 +91,7 @@ module Bootsnap
         FileUtils.mv(tmp, @store_path)
       rescue Errno::EEXIST
         retry
-      rescue Errno::EROFS
+      rescue SystemCallError
       end
     end
   end
--- a/lib/bootsnap/load_path_cache/store.rb
+++ b/lib/bootsnap/load_path_cache/store.rb
@@ -91,6 +91,7 @@ module Bootsnap
         FileUtils.mv(tmp, @store_path)
       rescue Errno::EEXIST
         retry
+      rescue Errno::EROFS
       end
     end
   end
--- a/test/load_path_cache/store_test.rb
+++ b/test/load_path_cache/store_test.rb
@@ -74,6 +74,12 @@ module Bootsnap
 
         store.transaction { store.set('a', 1) }
       end
+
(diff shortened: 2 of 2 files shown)


metaskills:
I did the CLA but I am not sure how to re-run the check.

lib/bootsnap/load_path_cache/store.rb
```diff
@@ -91,6 +91,7 @@ def dump_data
         FileUtils.mv(tmp, @store_path)
       rescue Errno::EEXIST
         retry
+      rescue Errno::EROFS
```
casperisfine, line 94: I'd prefer a more general solution. I think you can rescue `SystemCallError` instead.

If there is a file system permission issue or whatever, etc.

metaskills: Confirmed (https://www.honeybadger.io/blog/understanding-rubys-strange-errno-exceptions/) thanks and done.

casperisfine:
Looks good to me. Did you try this in your prod environment?

Also ultimately I'd suggest to do whatever is necessary so that your runtime `$LOAD_PATH` is the same as the build time one.

But either way this is a good fix IMHO, and I'll merge on green CI.

PR #361, opened 2021-05-05
Add prepend and append to ChangeObserver

psparrow:
This PR adds `prepend` and `append` to `ChangeObserver` to make bootsnap compatible with code that uses those methods to modify the `$LOAD_PATH`.

### DIFF
--- a/lib/bootsnap/load_path_cache/change_observer.rb
+++ b/lib/bootsnap/load_path_cache/change_observer.rb
@@ -15,11 +15,13 @@ module Bootsnap
           @lpc_observer.push_paths(self, *entries.map(&:to_s))
           super
         end
+        alias_method :append, :push
 
         def unshift(*entries)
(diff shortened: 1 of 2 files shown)


lib/bootsnap/load_path_cache/change_observer.rb
```diff
@@ -16,11 +16,21 @@ def push(*entries)
           super
         end
 
+        def append(*entries)
+          @lpc_observer.push_paths(self, *entries.map(&:to_s))
+          super
+        end
```
casperisfine, line 22: Why not just use `alias_method`?

casperisfine:
Looks like the test failure is legit.

psparrow:
> Looks like the test failure is legit.

Fixed. I moved the reinitialization test to a separate example for readability.

casperisfine:
Thanks.

Issue #243, opened 2019-03-08
Encoding::UndefinedConversionError when added Bootsnap (Rails 5.2.2 / Ruby 2.6.1).


bartekb:
Hi guys,

I have an issue when adding bootsnap into my app. My stack is based on Rails 5.2.2 / Ruby 2.6.1 / Ubuntu 16.04 / Puma 3.12.0 / Rubygems 3.0.3 / Bundler 1.17.3. I'm able to boot locally but when I open the app I get an Encoding::UndefinedConversionError. The app was upgraded from 4.2.11 to 5.1 and then 5.2.2. It works perfectly without bootsnap added. I've got the same error when deployed to Heroku (Ubuntu 18.04 stack). Below is a full framework traceback:

```
Encoding::UndefinedConversionError ("\xC3" from ASCII-8BIT to UTF-8):
  
activesupport (5.2.2) lib/active_support/core_ext/object/json.rb:38:in `encode'
activesupport (5.2.2) lib/active_support/core_ext/object/json.rb:38:in `to_json'
activesupport (5.2.2) lib/active_support/core_ext/object/json.rb:38:in `to_json'
activesupport (5.2.2) lib/active_support/json/encoding.rb:57:in `to_json'
json (2.2.0) lib/json/common.rb:224:in `generate'
json (2.2.0) lib/json/common.rb:224:in `generate'
activesupport (5.2.2) lib/active_support/json/encoding.rb:102:in `stringify'
activesupport (5.2.2) lib/active_support/json/encoding.rb:35:in `encode'
activesupport (5.2.2) lib/active_support/json/encoding.rb:22:in `encode'
activesupport (5.2.2) lib/active_support/core_ext/object/json.rb:41:in `to_json'
i18n-js (3.2.1) lib/i18n/js/formatters/base.rb:17:in `format_json'
i18n-js (3.2.1) lib/i18n/js/formatters/js.rb:10:in `block in format'
i18n-js (3.2.1) lib/i18n/js/formatters/js.rb:9:in `each'
i18n-js (3.2.1) lib/i18n/js/formatters/js.rb:9:in `format'
i18n-js (3.2.1) lib/i18n/js/segment.rb:51:in `write_file'
i18n-js (3.2.1) lib/i18n/js/segment.rb:38:in `save!'
i18n-js (3.2.1) lib/i18n/js.rb:45:in `each'
i18n-js (3.2.1) lib/i18n/js.rb:45:in `export'
[...]

PR #364, opened 2021-07-26
Set encoding for write pipe

ngan:
This fixes the `Encoding::UndefinedConversionError` error for Rails applications. Rails sets the default encoding to UTF8: https://github.com/rails/rails/blob/main/railties/lib/rails.rb#L22-L23

We want to set it to Binary but the IO.pipe method only sets it for the reader, not the writer.

This fixes https://github.com/Shopify/bootsnap/issues/243.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Fix `Encoding::UndefinedConversionError` error for Rails applications when precompiling cache. (#364)
+
 # 1.7.5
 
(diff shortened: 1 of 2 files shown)


ngan:
@byroot @casperisfine wdyt?

casperisfine (approved):

casperisfine:
Hum, CI is failing but for a totally unrelated reason (likely load order). I'll fix it on master.

casperisfine:
And thank you!

PR #372, opened 2021-09-22
Fixup `Kernel.load`

ojab:
It requires exact filename and doen't try any extensions, so currently
bootsnap implementation has two issues:
1. `load 'a'` will load `a.rb`
2. `load 'a'` will not load `a` unless `development_mode: true` and
fallback scan is activated

We can add this records to the cache, but probably it's too obscure
to justify more complex code and will not lead to improved
performance.

We can add all files to the store in `PathScanner#call`, but AFAIU
there is extension check there to skip probably non-requireable file
for performance reasons

### DIFF
--- a/lib/bootsnap/load_path_cache/cache.rb
+++ b/lib/bootsnap/load_path_cache/cache.rb
@@ -44,14 +44,20 @@ module Bootsnap
 
       # Try to resolve this feature to an absolute path without traversing the
       # loadpath.
-      def find(feature)
+      def find(feature, try_extensions: true)
         reinitialize if (@has_relative_paths && dir_changed?) || stale?
         feature = feature.to_s.freeze
+
         return feature if absolute_path?(feature)
-        return expand_path(feature) if feature.start_with?('./')
+
+        if feature.start_with?('./')
+          return try_extensions ? expand_path(feature) : File.expand_path(feature).freeze
+        end
+
         @mutex.synchronize do
-          x = search_index(feature)
+          x = search_index(feature, try_extensions: try_extensions)
           return x if x
+          return unless try_extensions
 
           # Ruby has some built-in features that require lies about.
           # For example, 'enumerator' is built in. If you require it, ruby
@@ -177,7 +183,9 @@ module Bootsnap
       end
 
       if DLEXT2
-        def search_index(f)
+        def search_index(f, try_extensions: true)
+          return try_index(f) unless try_extensions
+
           try_index(f + DOT_RB) || try_index(f + DLEXT) || try_index(f + DLEXT2) || try_index(f)
         end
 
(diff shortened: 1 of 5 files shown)


lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
```diff
@@ -56,27 +56,9 @@ def require_relative(path)
 
   alias_method(:load_without_bootsnap, :load)
   def load(path, wrap = false)
-    if (resolved = Bootsnap::LoadPathCache.load_path_cache.find(path))
-      return load_without_bootsnap(resolved, wrap)
-    end
-
-    # load also allows relative paths from pwd even when not in $:
```
ojab, line 63: `load_without_bootsnap` already doing it.

lib/bootsnap/load_path_cache/cache.rb
```diff
@@ -44,14 +44,19 @@ def load_dir(dir)
...
+
+        if feature.start_with?('./')
+          return try_extensions ? expand_path(feature) : File.expand_path(feature).freeze
+        end
+
         @mutex.synchronize do
-          x = search_index(feature)
-          return x if x
+          x = search_index(feature, try_extensions: try_extensions)
+          return x if x || !try_extensions
```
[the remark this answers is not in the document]
ojab: done.

lib/bootsnap/load_path_cache/cache.rb
```diff
@@ -44,14 +44,20 @@ def load_dir(dir)
...
+
+        if feature.start_with?('./')
+          return try_extensions ? expand_path(feature) : File.expand_path(feature).freeze
+        end
+
         @mutex.synchronize do
-          x = search_index(feature)
+          x = search_index(feature, try_extensions: try_extensions)
           return x if x
+          return x unless try_extensions
```
casperisfine, line 60: just `return unless try_extensions`. If you get there it means `x = nil`, so no point retuning `x`.

ojab: oh, right :woman_facepalming:

casperisfine:
This is a very good catch, and the patch look good to me beside minor styling issues. Big thanks!

casperisfine:
Also please add a CHANGLOG entry.

ojab:
Forgot to add a new file with tests for `Kernel.load`, pushed.

test/load_path_cache/core_ext/kernel_require_test.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+require('test_helper')
+
+module Bootsnap
+  module LoadPathCache
+    class CacheTest < MiniTest::Test
```
casperisfine, line 6:
```suggestion
    class KernelLoadTest < MiniTest::Test
```
The file should also probably be `kernel_load_test.rb`

ojab: copypaste is bad, fixed.

PR #373, opened 2021-09-22
Optimize relative `../` paths resolution

ojab:
`../` is also guaranteed to be relative path.


lib/bootsnap/load_path_cache/cache.rb
```diff
@@ -50,7 +50,7 @@ def find(feature, try_extensions: true)
 
         return feature if absolute_path?(feature)
 
-        if feature.start_with?('./')
+        if feature.start_with?('./') || feature.start_with?('../')
```
casperisfine, line 53:
```suggestion
        if feature.start_with?('./', '../')
```

ojab: TIL

PR #380, opened 2021-11-24
Only disable the compile cache for source files impacted by Ruby 3.0.3 [Bug #18250]

casperisfine:
I'm just kinda worried some source files may not explicitly break, but might load some corrupted iseqs instead.

I'll run this branch against our CI to make sure.


casperisfine:
It passed our internal CI just fine, so I'm confident we should be ok. I'll cut a new release.

PR #382, opened 2021-12-06
Update minimum version of msgpack

pravi:
With msgpack 1.1 tests fail with NoMethodError: undefined method `dump' for #<MessagePack::Factory:0x000055abe3562c70>

### DIFF
--- a/bootsnap.gemspec
+++ b/bootsnap.gemspec
@@ -44,5 +44,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency("minitest", "~> 5.0")
   spec.add_development_dependency("mocha", "~> 1.2")
 
-  spec.add_runtime_dependency("msgpack", "~> 1.0")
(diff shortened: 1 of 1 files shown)


pravi:
This will need dropping support for ruby 2.3 as msgpack now needs ruby >= 2.4

casperisfine:
Fixed in https://github.com/Shopify/bootsnap/commit/cd68c7c243b0bce2ef8cc7339fa8cdfeab75ba37.

Thanks.

casperisfine:
@kg-currenxie I'm sorry, I don't understand your question.

Do you mean that you have another dependency which specifically require `1.2` and no higher?

kg-currenxie:
it doesnt :) just that i have to add the gem specifically

`bootsnap` requires ~1.4 (which was the change in this PR)
a very popular gem i'd say :)

Issue #383, opened 2022-01-01
Kernel#require conflict with Zeitwerk


fxn:
Opening this issue to have a conversation about the conflict and its solution.

As you know, Zeitwerk decorates `Kernel#require` to detect autoloaded files it manages ([here](https://github.com/fxn/zeitwerk/blob/main/lib/zeitwerk/kernel.rb)).

There's an edge case where both libraries are in conflict: In order to load implicit namespaces on demand and have them behave like everything else, Zeitwerk issues an `autoload` call in one of the directories that implicitly define the namespace.

For example, if a Rails application has a directory

```
app/controllers/admin
```

and there is no file `admin.rb` defining the namespace, Zeitwerk issues

```ruby
Object.autoload(:Admin, "/absolute/path/to/app/controllers/admin")
```

When a `require` call for that directory is detected, Zeitwerk hijacks it, does not invoke the original `require`, and does not modify `$LOADED_FEATURES`. It just detects the implicit namespace being autoloaded, autovivifies the `Admin` module, sets autoloads for that namespace, invokes callbacks, logs if enabled, etc.

Now, here's the conflict: if Bootsnap's `require` is the one calling Zeitwerk's (depends on load order), Bootsnap's _assumes_ the directory is pushed to `$LOADED_FEATURES`. Bootsnap's loaded features cache gets out of sync, and reloads do not work because the cache does not reflect what is caching (it has an entry for the directory that never was in `$LOADED_FEATURES` and therefore was not removed from it).

I tested pushing the directory to `$LOADED_FEATURES`. That's like 4x slower in earlier Ruby versions, and in Ruby 3.1 the performance regression skyrockets to about 180x in [some tests I did](https://bugs.ruby-lang.org/issues/18452).

@jeremyevans commented in [ruby#18452](https://bugs.ruby-lang.org/issues/18452) that Bootsnap's caching logic might need to be revised.

Ideas:

* Let the cache reflect the true contents of `$LOADED_FEATURES` somehow.
* Let Bootsnap's `require` ignore absolute paths altogether. Zeitwerk exclusively uses absolute paths for everything.

What do you think?

byroot:
Ack. I'm still in vacation for another week, but I'll dig into this first thing when I get back to work.

We can likely ignore absolute path entirely yes, as a matter of fact Shopify has been using `config.add_autoload_paths_to_load_path = false` for a couple years now, because there's no point having load path caching for Zeitwerk managed files.

[...]

PR #385, opened 2022-01-10
LoadedFeaturesIndex#register: Ignore absolute paths entirely

casperisfine:
Fix: https://github.com/Shopify/bootsnap/issues/383
Ref: https://github.com/fxn/zeitwerk/issues/198

The loaded feature index goal is to prevent requiring the same "feature" twice if the `$LOAD_PATH` was mutated. e.g:

```ruby
require "bundler" # ~/gems/bundler/lib/bundler.rb
$LOAD_PATH.unshift("some/path")
require "bundler" # some/path/bundler.rb
```

In such scenario Ruby remember that you require "bundler" already and even though it now maps to a new path, it won't load it again.

This was causing issues with Zeitwerk if it is loaded before bootsnap (see refs), because the cache wouldn't be cleared.

But ultimately Zeitwerk always require absolute paths, and the concern described above simply doesn't apply to absolute paths. So we can simply bail out early in these cases. That fixes the bug, and also means less work and a smaller index, so win-win.

cc @fxn

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+* Ignore absolute paths in the loaded feature index. (#385)
+  This fixes a compatibility issue with Zeitwerk when Zeitwerk is loaded before bootsnap. It also should
+  reduce the memory usage and improve load performance of Zeitwerk managed files.
+
 * Automatically invalidate the load path cache whenever the Ruby version change. (#387)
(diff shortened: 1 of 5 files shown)


zarqman:
Not the original issue owner, but we saw this too and can confirm this PR fixes the issue here. Thanks!

fxn:
Tried and works as expected. Thanks @casperisfine ❤️.

casperisfine:
1.9.4 is out!

Issue #384, opened 2022-01-04
The same cache seems used for different Ruby versions


eregon:
Before https://github.com/Shopify/yjit-bench/pull/69#pullrequestreview-843427951 I and @noahgibbs noticed that Bootsnap has an invalid cache when running the `erubi_rails` benchmark with e.g. 3.1-dev, 3.1.0 and 3.0.x.
It seems the cache is not separated per Ruby version.

`rm -rf tmp/cache/bootsnap` is a workaround, but hard to remember, and the error can be [quite confusing](https://github.com/Shopify/yjit-metrics/issues/119#issuecomment-990812163).

Should bootsnap care about this?
It seems not so uncommon to use e.g. different Ruby releases to run a Rails app, and Rails seems to just use the default bootsnap cache path.

One issue is if CRuby dev versions are "supported" they report the same RUBY_VERSION even though many things including the ABI, the stdlib files, the bytecode format, etc might change. Maybe using RUBY_REVISION (the commit) is the only safe way then.

casperisfine:
👋 the ruby version is collected here: https://github.com/Shopify/bootsnap/blob/a0aa3d2b0cbdcef9d8db9ae787c3628ac250780f/ext/bootsnap/bootsnap.c#L246-L264

It's basically `RUBY_REVISION`. 

If you think there's a better key, I'm totally open to update it.

eregon:
Using `RUBY_REVISION` should be enough, although here a Hash of it is used.
What's the expected behavior if the `RUBY_REVISION` does not match when using bootsnap, e.g., in a Rails app?
I wonder if the issues we were hitting are maybe not guarded by that check/hash (e.g., the known stdlib files as in https://github.com/Shopify/yjit-metrics/issues/119#issuecomment-990812163).
Or maybe the hash of the revision isn't good enough, but that sounds not so likely.

casperisfine:
> What's the expected behavior if the RUBY_REVISION does not match when using bootsnap, e.g., in a Rails app?

The cache entry contain a "version" header. If the header doesn't match, it acts as a cache miss and is regenerated.

> I wonder if the issues we were hitting are maybe not guarded by that check/hash

I think you are right, that file was extracted in 3.1: https://github.com/ruby/ruby/commit/12a0a89e22fbc312e4a95a7749bc153532daa855

[...]

PR #387, opened 2022-01-10
Ignore load path caches generated by a different ruby version

casperisfine:
Fix: https://github.com/Shopify/bootsnap/issues/384

Until now it was assume that if the `$LOAD_PATH` was identical, then the content of the paths would be too.

However from one minor ruby version to another, the layout of the stdlib can change. So if the newer version is installed in the same place than the previous one, it might cause the load path cache to be invalid.

So we store `RUBY_DESCRIPTION` in the cache and make sure it matches before reusing the cache.

cc @eregon @noahgibbs if you mind giving this a quick look.


casperisfine:
Hum, I just remembered that what I liked too in `RUBY_DESCRIPTION` is that it contains the platform. E.g. if you try to use the same cache on your host and some kind of linux container and that somehow the ruby install path is the same (very unlikely, but might as well handle it). But I can use `RUBY_REVISION + RUBY_PLATFORM`

casperisfine:
Done. @eregon if you want to give it another look.

eregon:
Couple review questions:
* I don't see where the version is saved in the cache, just where it's read
* Is it OK to not use an existing cache from a previous Bootsnap version? Will it be overridden at the end of the run or so?

Issue #386, opened 2022-01-10
Investigate dropping the load path cache at the end of boot 


casperisfine:
Something I wanted to do for a while and that @SamSaffron reminded me of: https://bugs.ruby-lang.org/issues/18454#note-5

For big applications, the load path cache can be quite big. But in production mode, most apps shouldn't need it once the boot sequence is over. 

So we should look at freeing it at the end of eager load (pre-fork).

SamSaffron:
I think just providing an API for releasing the index is probably fine as a first step, cause we could just wire it into our last Rails initializer. 

`Bootsnap.dispose_index` perhaps?

casperisfine:
Yes that's trivial. But I'd like to see if we could provide a `Railtie` to do this automatically so that all users benefit from it and not just a tiny minority. That's why I need to dig a bit more.

PR #388, opened 2022-01-11
Add Bootsnap.unload_cache!

casperisfine:
Fix: https://github.com/Shopify/bootsnap/issues/386

I tested it on our monolith, seem to work and frees about 7.4MB of memory (it's really a big app, so it would be much less for the average user).

Not quite sure what's the best way to integrate it is just yet. Right now I've put:

```ruby
# config/application.rb
    config.after_initialize do
      if config.eager_load
        puts "[Bootsnap] unload cache"
        Bootsnap.unload_cache!
      end
    end
```

But it might make sense to only do this in `config/environments/production.rb`, because we do have eagerloading on CI and this would trigger before the tests and `test_helper.rb` are loaded, so it won't be able to help loading test dependencies. Also Memory usage is generally much less critical on CI.

As said on the issue it would be nice to make this happen automatically with a Rails of some sort, but it might be tricky given how early bootsnap is loaded. Maybe through a separate `bootsnap-rails` gem?

cc @SamSaffron

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,6 +27,8 @@ jobs:
       - run: bundle exec rake
 
   rubocop:
+    strategy:
+      fail-fast: false
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -38,6 +40,7 @@ jobs:
 
   rubies:
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu]
         ruby: ['2.5', '2.6', '2.7', '3.0', '3.1', 'ruby-head', 'debug']
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
   People still using the classic autoloader might want to stick to `bootsnap 1.12`.
 
+* Add `Bootsnap.unload_cache!`. Applications can call it at the end of their boot sequence when they know
+  no more code will be loaded to reclaim a bit of memory.
+
 # 1.12.0
 
 * `bootsnap precompile` CLI will now also precompile `Rakefile` and `.rake` files.
--- a/lib/bootsnap.rb
+++ b/lib/bootsnap.rb
@@ -76,6 +76,10 @@ module Bootsnap
       )
     end
 
+    def self.unload_cache!
+      LoadPathCache.unload!
+    end
+
     def iseq_cache_supported?
       return @iseq_cache_supported if defined? @iseq_cache_supported
 
--- a/lib/bootsnap/load_path_cache.rb
+++ b/lib/bootsnap/load_path_cache.rb
@@ -21,8 +21,12 @@ module Bootsnap
 
     CACHED_EXTENSIONS = DLEXT2 ? [DOT_RB, DLEXT, DLEXT2] : [DOT_RB, DLEXT]
 
+    @enabled = false
+
     class << self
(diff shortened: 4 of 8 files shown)


rafaelfranca:
> Maybe through a separate `bootsnap-rails` gem?

Would be a problem to just add a Railtie to bootsnap? Maybe requiring Rails before the bootsnap cache is setup could be a problem, but we could only load the Railtie in the last line of `bootsnap/setup.rb`.

Or add a new `bootsnap/rails` file that load `setup` and define the `Railtie`. Inside the `Railtie` we can add the after initialize that only run in production by default, but can be controlled with a `config.bootsnap.unload_cache` in case people have a staging environment.

casperisfine:
> The benches use of RSS is not consistent

As said on https://bugs.ruby-lang.org/issues/18454#note-5, RSS isn't a good benchmark metric for this. What we do here is release objects at the end of boot. That makes space in the Ruby heap for other objects to use it, but Ruby won't release that memory to the system so RSS won't be reduced. You'd need to call `malloc_trim` or similar for that (I think jemalloc can be configured to release memory more aggressively too?)

Also note that this happens pre-fork, so that's memory that would have been in CoW for the most part, so depending on how many workers you have it may not be that significant. 

If you want to measure the difference you should use heap-profiler / memory-profiler.

casperisfine:
> Would be a problem to just add a Railtie to bootsnap? Maybe requiring Rails before the bootsnap cache is setup could be a problem, but we could only load the Railtie in the last line of bootsnap/setup.rb.

The problem is that `bootsnap/setup` doesn't assume Rails, and since it's require from `config/boot.rb` `Rails` isn't defined yet, so we can't use the usual `require "railtie" if defined? ...`.

> Or add a new bootsnap/rails file that load setup and define the Railtie

That make more sense, but is it ok to load the code required for railties earlier than before?

rafaelfranca:
> That make more sense, but is it ok to load the code required for railties earlier than before?

Not really since bundler is not even setup yet when we load bootsnap. I think I'd be ok to add the railtie require inside `config/application.rb` by default though, but it is possible DHH will complain 😂.

rafaelfranca:
f we add the railtie require inside `bootsnap.rb` it would be fine since that file would be required in `config/application.rb`. We would have to move the content of that file to some `bootsnap/core` and keep it being just:

```
require "bootsnap/core"
require "bootsnap/rails" if defined?(Rails::Railtie)
```

I think that would work and would not require any app to change their code.

ngan:
This is really cool! Imho I’m really digging the separate `bootsnap-rails` gem path. If the goal is to make bootsnap rails agnostic, that would separate the concerns. We can have the new gem list this gem as a dependency but not require it. This means we can change rails default Gemfile to install the new gem without the funky `require: false` and the comment to explain why. The boot file will continue to do the bootsnap/setup require. As for rollout, we can add this unload method and it’ll pretty much do nothing and just tell people to switch to bootsnap-rails if they’re on rails.

rafaelfranca:
Bootsnap is already Rails agnostic. That is not the goal of this change. I'm also strong against adding a new gem. We can solve this without new gems, and if anything make bootsnap less Rails agnostic.

ngan:
> Bootsnap has no goals of being Rails agnostic

Ah! Makes sense

> `require: false` we actually don't need that

Maybe? https://github.com/rails/rails/pull/44267

PR #389, opened 2022-01-12
Refactor core_ext to use prepend over alias method chains

casperisfine:
Ref: https://github.com/Shopify/bootsnap/pull/388#issuecomment-1010002918

Alias method chain makes it very hard to cleanly unhook, because other gems like Zeitwerk might have done the same, so we can't just alias back, as it may replace a method that wasn't our own.

No idea if this could actually work, I need to test it more.

### DIFF
--- a/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
+++ b/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
@@ -1,65 +1,76 @@
 # frozen_string_literal: true
 
-module Kernel
-  module_function
-
-  alias_method(:require_without_bootsnap, :require)
-
-  def require(path)
-    string_path = Bootsnap.rb_get_path(path)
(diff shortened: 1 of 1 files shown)


casperisfine:
Hum, according to Zeitwerk comments this won't work: https://github.com/Shopify/zeitwerk/blob/f785c89cf4f6111552c1ab1f2e0346cc37a5a085/lib/zeitwerk/kernel.rb#L18-L20

casperisfine:
Ok, so I tested it with the following script:

```
module KernelRequire
  def require(path)
    p(path)
    super
  end
end

Kernel.prepend(KernelRequire)
# Kernel.singleton_class.prepend(KernelRequire)

p 1
require 'net/http'

p 2
Kernel.require('prime')

p 3
class Foo
  require 'objspace'
end
```

It work with Ruby 3.0+, but not with 2.7 and older :/

PR #390, opened 2022-01-12
Updates

gmcgibbon:
Update license, default branch references, and rubocop.

cc @Shopify/rails


adrianna-chang-shopify (approved):

.rubocop.yml
```diff
@@ -1,11 +1,13 @@
...
+
+inherit_gem:
+  rubocop-shopify: rubocop.yml
 
 AllCops:
   Exclude:
     - 'vendor/**/*'
     - 'tmp/**/*'
-  TargetRubyVersion: '2.4'
+  TargetRubyVersion: 2.5
```
casperisfine, line 10: Hum, we still support Ruby 2.3

gmcgibbon: We do, but since it is EOL and parser doesn't include support for it anymore we should probably up the min ruby version if it isn't too much trouble.

casperisfine: I'd rather drop the dependency on the company style guide and use an older parser/rubocop.

Issue #375, opened 2021-09-28
Add caching for YAML.safe_load_file


jesse-shopify:
### I made sure the issue is in bootsnap

I learned today that files loaded by YAML.safe_load_file are not cached. Searching this repository suggests this is the case with YAML.load_file and YAML.unsafe_load_file being supported but not YAML.safe_load_file.

### Steps to reproduce

The application where this is relevant is using the Faker gem, thus stuck at Psych 3.3. To use safe_load_file in Psych 3.3 it has to be called explicitly (in Psych 4.0 it is the default for load_file).

### Expected behavior

Files loaded with YAML.safe_load_file should be cached (at least for simple use cases).

### Actual behavior

Files loaded with YAML.safe_load_file are not cached.

### System configuration

**Bootsnap version**: 1.9.1

**Ruby version**: 2.6.3, 2.7.1

**Rails version**: 6.1.4.1

casperisfine:
It wasn't done because to be efficient, the caching must be doable statically through the precompile command, so at this stage we have no idea what kind of arguments will be used. 

So at best we could support safe loading only when a strict subset of arguments are used.

casperisfine:
Arf, I remember now why I didn't do it. 

For options such as `alias: false`, that information is lost once the cache is compiled, so for instance:

```ruby
YAML.safe_load_file("foo.yml") # raises because the file has aliases or other unsafe constructs
YAML.unsafe_load_file("foo.yml") # works fine and the cache is stored
YAML.safe_load_file("foo.yml") # would succeed because it would load the cache directly.
```

I really don't see any way around this.

jesse-shopify:
Isn't this already an issue with Psych 4 since ```load_file``` calls ```load``` which calls ```safe_load```?
- https://github.com/ruby/psych/blob/master/lib/psych.rb#L669
- https://github.com/ruby/psych/blob/master/lib/psych.rb#L369

If I'm reading this right, the only possible supported options are ```symbolize_keys``` and ```freeze``, any other option will not hit the cache (also good to know):
- https://github.com/Shopify/bootsnap/blob/master/lib/bootsnap/compile_cache/yaml.rb#L124

If that's the case then passing ```alias``` will always avoid the cache. An implementation of ```safe_load_file``` identical to ```load_file``` and ```unsafe_load_file``` seems like it would be sufficient for the common use case.

[...]

PR #392, opened 2022-01-13
Handle YAML.load_file on Psych 4

casperisfine:
Kind of a fix for https://github.com/Shopify/bootsnap/issues/375. Not exactly because it doesn't add support for `safe_load_file`, but for `YAML.load_file` on `psych 4`, which is the most used interface.

The idea is that we store a boolean prefix in the cache. When we precompile we try to parse in safe mode, if it fails we fallback to unsafe parsing.

When loading from cache, we know if `load_file` or `unsafe_load_file` is used. `unsafe_load_file` can use cache generated in safe mode, but `load_file` won't use caches generated via `unsafe_load_file`.

This is a rough prototype, it misses a few bits and needs a lot more testing.


casperisfine:
Ok, found an interesting bug:

```ruby
File.write("foo.json", "true")
JSON.load_file("foo.json) # => true
YAML.load_file("foo.json) # EOFError
```

It might seem weird, but since YAML is a superset of JSON, it's not totally invalid, and even if it was we should still fallback properly.

Until now it worked because JSON and YAML caches had the same format. But now that YAML caches have a prefix to indicate wether they are safe mode compatible, YAML can't read JSON caches and vice-versa. 

So each cache would need a key prefix or something.

casperisfine:
The Windows segfault seem consistent and legit.

PR #396, opened 2022-01-21
Respect `Kernel#require`'s duck-type

etiennebarrie:
`Kernel#require` turns its argument into a path by calling `to_path` if it responds to it, and then into a String (the argument or the return value of `to_path`) by using implicit conversion (`to_str`).

We can see this in the spec: https://github.com/ruby/spec/blob/97364ff1f90e63ba0159216f9563318a0d97dceb/core/kernel/shared/require.rb#L51

* it calls #to_str on non-String objects
* it calls #to_path on non-String objects
* it calls #to_str on non-String objects returned by #to_path

Or in MRI, in `rb_get_path`, called from `require_internal`:
https://github.com/ruby/ruby/blob/5e3a32021849718ae483eaaa9fbf155f91828039/file.c#L219-L221

---

I used a fork because I noticed none of the tests were loading the core ext (even the test below this one supposedly testing `load`, I'll follow-up on that) and wanted to avoid leaks for other tests.

### DIFF
--- a/test/load_path_cache/core_ext/kernel_require_test.rb
+++ b/test/load_path_cache/core_ext/kernel_require_test.rb
@@ -4,6 +4,31 @@ require("test_helper")
 
 module Bootsnap
   module KernelRequireTest
+    class KernelRequireTest < Minitest::Test
+      def test_uses_the_same_duck_type_as_require
+        assert_nil LoadPathCache.load_path_cache
+        cache = Tempfile.new("cache")
+        pid = fork do
+          LoadPathCache.setup(cache_path: cache, development_mode: true)
+          dir = File.realpath(Dir.mktmpdir)
+          $LOAD_PATH.push(dir)
+          FileUtils.touch("#{dir}/a.rb")
+          stringish = mock
+          stringish.expects(:to_str).returns("a").twice # bootsnap + ruby
+          pathish = mock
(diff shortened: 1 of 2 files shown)


test/load_path_cache/core_ext/kernel_require_test.rb
```diff
@@ -4,6 +4,31 @@
...
+          stringish = mock
+          stringish.expects(:to_str).returns("a").twice # bootsnap + ruby
+          pathish = mock
+          pathish.expects(:to_path).returns(stringish).twice # bootsnap + ruby
+          assert pathish.respond_to?(:to_path)
+          require(pathish)
+          FileUtils.rm_rf(dir)
+        end
+        Process.wait(pid)
+        assert_predicate Process.last_status, :success?
```
casperisfine, line 25: Ruby 2.3 compat

```suggestion
        pid, status = Process.wait2(pid)
        assert_predicate status, :success?
```

casperisfine:
Silly idea, `rb_get_path` is public API, and since we have a C ext already, we could just expose it on `Bootsnap::Native` and use it.

casperisfine:
here's how it could look like https://github.com/Shopify/bootsnap/commit/17ea17b3a4a11e907225e2dc3f972eb0e0d66dbf

PR #399, opened 2022-01-31
Simplified handling of UTF-8 symbols in YAML

casperisfine:
We actually don't need to store the encoding, we can assume
everything is UTF-8. `Symbol#to_sym` will take care of coercing
to ASCII when applicable.

Any other encoding is not supported as the cache should be invoked
if Psych could generate any other encoding.

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -2,7 +2,7 @@
  * Suggested reading order:
  * 1. Skim Init_bootsnap
  * 2. Skim bs_fetch
- * 3. The rest of everything
+ * 3. The rest of everyrything
  *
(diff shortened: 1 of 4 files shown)


ext/bootsnap/bootsnap.c
```diff
@@ -2,7 +2,7 @@
  * Suggested reading order:
  * 1. Skim Init_bootsnap
  * 2. Skim bs_fetch
- * 3. The rest of everything
+ * 3. The rest of everyrything
```
teoljungberg, line 5: I presume this was typed by accident: `everything`?

teoljungberg: Opened as https://github.com/Shopify/bootsnap/pull/405

PR #401, opened 2022-02-04
Improve handling of chaining #require_relative patch

harrycis:
Kernel.require_relative could be patched again by other Gems initialized after bootsnap. This change supports the chaining of Kernel.require_relative

### DIFF
--- a/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
+++ b/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
@@ -35,7 +35,7 @@ module Kernel
 
   alias_method(:require_relative_without_bootsnap, :require_relative)
   def require_relative(path)
-    location = caller_locations(1..1).first
+    location = caller_locations.find { |x| x.base_label != "require_relative" }
     realpath = Bootsnap::LoadPathCache.realpath_cache.call(
       location.absolute_path || location.path, path
     )
--- a/test/load_path_cache/core_ext/kernel_require_relative_test.rb
+++ b/test/load_path_cache/core_ext/kernel_require_relative_test.rb
@@ -33,6 +33,7 @@ module Bootsnap
         FileUtils.rm_rf(@dir1)
         Kernel.module_eval do
           module_function
+
           undef :require_relative
           alias_method :require_relative, :pre_patch_require_relative
           undef :pre_patch_require_relative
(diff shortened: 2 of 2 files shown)


lib/bootsnap/load_path_cache/core_ext/kernel_require.rb
```diff
@@ -35,7 +35,7 @@ def require(path)
 
   alias_method(:require_relative_without_bootsnap, :require_relative)
   def require_relative(path)
-    location = caller_locations(1..1).first
+    location = caller_locations.find { |x| x.base_label != "require_relative" }
```
byroot, line 38: Hum this increase the cost quite significantly.

Out of curiosity what else decorates `require_relative`? Maybe we could do something like `caller_locations(1..3)` to support up to two extra decorators?

harrycis: Honestly, not aware of any public Gem decorate it, but our projects internally is about to decorate it to track some state change pre/post file load, and it's blocked by the decorator in bootnap. 

Only traverse the first few caller seems fine to me. The chaining of require_relative may not happen that often probably.

byroot: I'll have to look on Monday if we even still need to decorate `require_relative`. I think it we might be able to remove this decorator entirely.

casperisfine: Ok, so this is the change I remembered that may allow us to remove the `require_relative` decorator: https://github.com/ruby/ruby/commit/79a4484a072e9769b603e7b4fbdb15b1d7eccb15. It shipped with Ruby 3.1, so that wouldn't remove it for everyone, but I'm much less concerned about a small perf impact if it doesn't apply to the newest rubies.

The decorator was added in https://github.com/Shopify/bootsnap/pull/136, I'll have to dig more to see what it would take to remove it on 3.1+.

casperisfine: Hum, the more I dig into https://github.com/Shopify/bootsnap/pull/136, the more it seems like we tried to limit a Ruby bug more than a Bootsnap bug.

Assuming `a/test.rb` and `b/test.rb` symlink each others, the following load the file twice on all rubies but 3.1:

```ruby
require_relative "b/test.rb"
require_relative "a/test.rb"
```

So I don't think we should have done this at all.

harrycis:
That would work. I am closing the PR then.

Thanks for the investigation, folks!

casperisfine:
Welcome. I might not release this right away as there's a couple other things I'd like to include. I recommend using the main branch in your Gemfile if you need a fix right away.

Issue #409, opened 2022-03-07
Cache key when using musl libc (eg Alpine linux)


dhagher:
When not using glibc, cache key is based on `uname -v` leading to the same problems when bundling cache as described in https://github.com/Shopify/bootsnap/pull/283
This makes bootsnap significantly less useful for container workloads when [Alpine Linux](https://alpinelinux.org) is used.

### Steps to reproduce
Run on non glibc system, eg Alpine Linux using docker
```FROM ruby:3.1.1-alpine3.15```

### Expected behavior
Cache precompiled and stored in the container, eg
```RUN bootsnap precompile --gemfile```
should be valid on other hosts.

### Actual behavior
Due to lack of `__GLIBC__`,  cache key depends on kernel version https://github.com/Shopify/bootsnap/blob/main/ext/bootsnap/bootsnap.c#L294
Based on https://wiki.musl-libc.org/faq.html#Q:-Why-is-there-no-%3Ccode%3E__MUSL__%3C/code%3E-macro? I think there is no similar way to check for musl version.

### System configuration

**Bootsnap version**: 1.10.3

**Ruby version**: 3.1.1

**Comment by casperisfine** [comment]
Hum, to be honest I don't quite get why we use such a precise cache key.

At first sight `RUBY_DESCRIPTION` should plenty precise for instruction sequences and load path caches but I might be wrong.

Git blame point to https://github.com/Shopify/bootsnap/pull/43:

```
 * The idea here is that we want a cache key member that changes when the OS
 * changes in such a way as to make existing compiled ISeqs unloadable.
 *
 * On Darwin and FreeBSD, this is the kernel version, since that tends to vary
 * with whole-system upgrades. What we probably care about more is the libc
 * version, which is what we explicitly ask for on linux.
 * (and KERN_OSRELEASE came back empty for me one one linux box, so...?)
 *
 * I'm kind of guessing about the important factors here. We could probably do
 * this better.
```

I'll try to consult a few people to see if we could drop theses from the cache key.

casperisfine:
From a colleague:

> The code bit casts ints and stuff, so endianness matters and maybe different on 32b versus 64b as well.

But I don't think `__GLIBC__ ` actually covers that.

dhagher:
for checking endianness in ruby I see suggestions to compare `pack("I")` with `pack("N")` which translates to a runtime cast check https://github.com/ruby/ruby/blob/ruby_3_1/pack.c#L59

[...]

PR #412, opened 2022-03-09
Reduce the cache key to RUBY_REVISION + RUBY_PLATFORM

casperisfine:
Fix: https://github.com/Shopify/bootsnap/issues/409

Various system versions were added, but as far as I could gather it was just overly conservative.

This was reducing the usefulness of bootsnap when using Alpine linux containers.

cc @dhagher: would you mind testing this branch with your app?


casperisfine:
> (can those also be precompiled like .rb files?)

Yes good point, I need to update a pattern in the CLI. 

> (btw function comment and readme refers to uname and might need updating with this change)

Good point too, I'll fix that and add a CHANGELOG entry.

PR #414, opened 2022-05-01
test_helper: support platforms where GC compaction is not supported

terceiro:
Not all platforms support GC compaction. On those, calling
CG.verify_compaction_references raises a NotImplementedError exception.


casperisfine (approved):
Thanks.

Unfortunately you'll have to sign the CLA.

terceiro:
> Unfortunately you'll have to sign the CLA.

done.

PR #424, opened 2022-11-08
Add a way to skip direcftories during load path scanning

casperisfine:
If you have large non-ruby directories in the middle of your load path, it can severly slow down scanning.
Typically this is a problem with `node_modules`

Fix: https://github.com/Shopify/bootsnap/issues/277

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+* Add a way to skip direcftories during load path scanning.
+  If you have large non-ruby directories in the middle of your load path, it can severly slow down scanning.
+  Typically this is a problem with `node_modules`. See #277.
 * Fix `Bootsnap.unload_cache!`, it simply wouldn't work at all becaue of a merge mistake. See #421.
 
 # 1.13.0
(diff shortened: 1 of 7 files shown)


bogn83:
NICE:
```bash
$ (time rails runner "puts 'b'") 2>&1 > /dev/null | grep real
real    0m4,661s
$ (time rails runner "puts 'b'") 2>&1 > /dev/null | grep real
real    0m4,525s
$ (time rails runner "puts 'b'") 2>&1 > /dev/null | grep real
real    0m4,588s
$ (time rails runner "puts 'b'") 2>&1 > /dev/null | grep real
real    0m4,691s
$ echo switch # pulled the PR in with a different terminal, this is just a marker
switch
$ (time rails runner "puts 'b'") 2>&1 > /dev/null | grep real
real    0m0,850s
$ (time rails runner "puts 'b'") 2>&1 > /dev/null | grep real
real    0m0,866s
$ (time rails runner "puts 'b'") 2>&1 > /dev/null | grep real
real    0m0,930s
$ (time rails runner "puts 'b'") 2>&1 > /dev/null | grep real
real    0m0,935s
$ (time rails runner "puts 'b'") 2>&1 > /dev/null | grep real
real    0m0,900s
```

Issue #423, opened 2022-11-03
Implement read only mode


ChrisBr:
Implement a read only mode which does not update the bootsnap cache anymore e.g by supplying an env variable `BOOTSNAP_READ_ONLY`.

## Background

Our Bootsnap cache in CI is quite significant and we run several CI workers on the same machine via docker. As bootsnap keeps writing to the cache, we need to keep a dedicated copy of the cache for each of our CI worker instances. If bootsnap would support a read only mode, we could only keep one copy of the cache around and mount it into the containers in read only mode. This would reduce our network traffic, I/o and disk size significantly.

casperisfine:
Ack. I think that would make sense, even beyond the use case you mention.

It may take me a tiny while before I can look into this (couple weeks perhaps).

PR #428, opened 2022-11-24
Add readonly mode

casperisfine:
Fix: https://github.com/Shopify/bootsnap/issues/423

When you know that the cache won't be used again, it avoid some useless work and IOs.

Typically this might be the case for dockerized applications. You generate a cache when building the image, but then when you boot the application any cache update won't be persisted, so thre is no point.


casperisfine:
I can cut a release easily yeah, but would be good if you double checked the use case you had works as expected. You can point the gemfile at the repo.

PR #432, opened 2023-01-12
Treat only StdLib Ruby files as stable.

### LINKED ISSUE #431: /usr/share/ruby/time.rb is not `.stable?` - path_test.rb failing
This is basically dupe of #173, but since it was closed without resolution, I am to going to file this issue again.

The problem is this failing test:

~~~
  1) Error:
Bootsnap::LoadPathCache::PathTest#test_stability:
NoMethodError: undefined method `pry' for #<Binding:0x00007f34f6608c90>
    /builddir/build/BUILD/bootsnap-1.15.0/usr/share/gems/gems/bootsnap-1.15.0/test/load_path_cache/path_test.rb:25:in `test_stability'
~~~

And I think the issue is in this line:

https://github.com/Shopify/bootsnap/blob/b6107b8c16ed6a0748386bd8f409f0f944eac767/lib/bootsnap/load_path_cache/path.rb#L128

So the path in question is:

~~~
[1] pry(#<Bootsnap::LoadPathCache::PathTest>)> p stable
#<Bootsnap::LoadPathCache::Path:0x00007fb3fa26dfa8 @path="/usr/share/ruby/time.rb", @real=false>
=> #<Bootsnap::LoadPathCache::Path:0x00007fb3fa26dfa8 @path="/usr/share/ruby/time.rb", @real=false>
~~~

while these are the paths used to match:

~~~
[2] pry(#<Bootsnap::LoadPathCache::PathTest>)> p RbConfig::CONFIG["libdir"]
"/usr/lib64"
=> "/usr/lib64"
[3] pry(#<Bootsnap::LoadPathCache::PathTest>)> p RbConfig::CONFIG["sitedir"]
"/usr/local/share/ruby/site_ruby"
=> "/usr/local/share/ruby/site_ruby"
~~~

The issue here is that the Ruby configuration allows more flexibility then the code assumes, therefore the file in question is wrongly marked as instable. That could probably lead to some other wrong assumptions made by Bootsnap, dunno.

voxik:
This fixes test failure on Fedora:

~~~
  1) Error:
Bootsnap::LoadPathCache::PathTest#test_stability:
NoMethodError: undefined method `pry' for #<Binding:0x00007f34f6608c90>
    /builddir/build/BUILD/bootsnap-1.15.0/usr/share/gems/gems/bootsnap-1.15.0/test/load_path_cache/path_test.rb:25:in `test_stability'
~~~

where the culprit is that Ruby StdLib lives in `/usr/share`, instead of `/usr/lib64`. However, this is the flexibility given by Ruby configuration, so this should be correctly handled.

But it seems, that the real intention always was to consider just Ruby StdLib as stable, because otherwise, there would not be exception for Site dir.

Therefore, cover just the StdLib via `RbConfig::CONFIG["rubylibdir"]` and treat everything else as unstable.

Please note that the arch specific libraries are not Bootsnap concern, therefore `rubyarchdir` is ignored.

Fixes #431

### DIFF
--- a/lib/bootsnap/load_path_cache/path.rb
+++ b/lib/bootsnap/load_path_cache/path.rb
@@ -115,17 +115,16 @@ module Bootsnap
       STABLE   = :stable
       VOLATILE = :volatile
 
-      # Built-in ruby lib stuff doesn't change, but things can occasionally be
-      # installed into sitedir, which generally lives under libdir.
-      RUBY_LIBDIR  = RbConfig::CONFIG["libdir"]
-      RUBY_SITEDIR = RbConfig::CONFIG["sitedir"]
+      # Condiser Ruby StdLib dir to be stable.
+      RUBY_LIBDIR = RbConfig::CONFIG["rubylibdir"]
+      RUBY_ARCHDIR = RbConfig::CONFIG["rubyarchdir"]
 
       def stability
         @stability ||= if Gem.path.detect { |p| expanded_path.start_with?(p.to_s) }
           STABLE
         elsif Bootsnap.bundler? && expanded_path.start_with?(Bundler.bundle_path.to_s)
           STABLE
-        elsif expanded_path.start_with?(RUBY_LIBDIR) && !expanded_path.start_with?(RUBY_SITEDIR)
+        elsif expanded_path.start_with?(RUBY_LIBDIR) || expanded_path.start_with?(RUBY_ARCHDIR)
           STABLE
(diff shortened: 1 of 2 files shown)


casperisfine:
Closing in favor of https://github.com/Shopify/bootsnap/commit/72202aab5e5b3602ece4e8748bcdeefe2d789ab5

I don't think the `ARCH_DIR` part was correct anyway.

PR #437, opened 2023-03-14
Remove extraneous memory allocations in C

ianks:
This PR makes use of `rb_str_buf_new`, so we can `read` directly into it. This avoids an an extraneous `ALLOC_N` and corresponding `xfree` for every item that is read.

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -467,7 +467,6 @@ open_cache_file(const char * path, struct bs_cache_key * key, const char ** errn
 static int
 fetch_cached_data(int fd, ssize_t data_size, VALUE handler, VALUE args, VALUE * output_data, int * exception_tag, const char ** errno_provenance)
 {
-  char * data = NULL;
   ssize_t nread;
   int ret;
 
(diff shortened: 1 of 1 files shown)


ext/bootsnap/bootsnap.c
```diff
@@ -713,8 +712,8 @@ bs_fetch(char * path, VALUE path_v, char * cache_path, VALUE handler, VALUE args
     else if (res == CACHE_UNCOMPILABLE) {
       /* If fetch_cached_data returned `Uncompilable` we fallback to `input_to_output`
         This happens if we have say, an unsafe YAML cache, but try to load it in safe mode */
-      if (bs_read_contents(current_fd, current_key.size, &contents, &errno_provenance) < 0) goto fail_errno;
-      input_data = rb_str_new(contents, current_key.size);
+      input_data = rb_str_buf_new(cached_key.size);
+      if (bs_read_contents(current_fd, current_key.size, input_data, &errno_provenance) < 0) goto fail_errno;
```
casperisfine, line 716: Since the return value of `bs_read_contents` is always checked against `< 0`, you could move `rb_str_buf_new` inside, and return `Qfalse` on failure.

ianks: Ok, made those adjustments in https://github.com/Shopify/bootsnap/pull/437/commits/d1c01f6a6f83d2cd307624be5085eecac3d8b5ad

ianks:
Should I handle changelog / merging?

PR #451, opened 2023-07-21
Remove custom permission error code and restore path in read error

rwstauner:
The existing error message says

> bootsnap doesn't have permission to write cache entries in '#{cache_path}' 
> (or, less likely, doesn't have permission to read '#{path}')"

but at this point i think the error will never be about the cache and only for the original file.

We stopped reporting write errors in
d878622782f8abed46afcbff1385af075c115859
and read errors in
503e9d50805769e9fc5034ed175062810e8f8f54

In that case it seems like we may as well remove the wrapper class and let the original error bubble up...
but when I did that the exception message was the code "bs_fetch:open_current_file:open" rather than mentioning what file path failed to be read.

My C is incredibly rusty; I don't know if this is a good idea.

What do you think? Do we want to do it this way?
Are there other places we should replace the "bs:*" code with the file path (like the fstat that follows?)

Another option would be to keep the ruby method and just simplify the message
(in which case the C code wouldn't have to change).

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -682,10 +682,14 @@ bs_fetch(char * path, VALUE path_v, char * cache_path, VALUE handler, VALUE args
   VALUE output_data;  /* return data, e.g. ruby hash or loaded iseq */
 
   VALUE exception; /* ruby exception object to raise instead of returning */
+  VALUE exception_message; /* ruby exception string to use instead of errno_provenance */
 
   /* Open the source file and generate a cache key for it */
   current_fd = open_current_file(path, &current_key, &errno_provenance);
-  if (current_fd < 0) goto fail_errno;
+  if (current_fd < 0) {
+    exception_message = path_v;
+    goto fail_errno;
+  }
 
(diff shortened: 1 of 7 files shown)


ext/bootsnap/bootsnap.c
```diff
@@ -368,7 +368,8 @@ open_current_file(char * path, struct bs_cache_key * key, const char ** errno_pr
 
   fd = open(path, O_RDONLY);
   if (fd < 0) {
-    *errno_provenance = "bs_fetch:open_current_file:open";
+    // ensure error for reading the input file contains the original path
+    *errno_provenance = path;
```
casperisfine, line 372: So `open_current_file` is only called from `bs_fetch`, and `path` here is `RSTRING_PTR` of the `path_v` `bs_fetch` received.

If we return `fd <0`, `bs_fetch` jumps to the `fail_errno` label, so we're not actually giving more information to the caller here, since it already have `path_v` on its stack.

If we wish to augment or change the error message, we can directly do it there.

casperisfine:
I like it, thanks.

PR #452, opened 2023-08-03
Fix compatibility with Minitest 5.19+

voxik:
The `MiniTest` was renamed to `Minitest`:

https://github.com/minitest/minitest/commit/9a57c520ceac76abfe6105866f8548a94eb357b6

And the `MiniTest` constant is now loaded just when `MT_COMPAT` environment variable is set:

https://github.com/minitest/minitest/commit/a2c6c18570f6f0a1bf6af70fe3b6d9599a13fdd6

### DIFF
--- a/test/helper_test.rb
+++ b/test/helper_test.rb
@@ -2,7 +2,7 @@
 
 require("test_helper")
 
-class HelperTest < MiniTest::Test
+class HelperTest < Minitest::Test
   include(CompileCacheISeqHelper)
   include(TmpdirHelper)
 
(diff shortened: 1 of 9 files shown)


voxik:
As always, I cannot sign the CLA, so do whatever you need to do with this PR (and I would not mind if you dropped the CLA)

casperisfine:
The CLA annoys me as much as you but unfortunately my hands are tied.

The test suite doesn't pass anyway, so I'll consider this PR as a bug report and fix it myself. Thanks.

Issue #474, opened 2024-01-31
EOFError: end of buffer reached


kuahyeow:
### I made sure the issue is in bootsnap

After disabling bootsnap, the issue now longer appears

### Steps to reproduce

Unfortunately, this only happens reliably with `gdk update` (GitLab tooling). I am working to find a simpler reproduction in https://gitlab.com/gitlab-org/gitlab/-/merge_requests/143230#note_1750589800 

### Actual behavior

The previous version of bootsnap we used as 1.17.0. After the upgrade to 1.18.1, `bundle exec rails db:migrate` started failing with:

```
rake aborted!
EOFError: end of buffer reached
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database/dictionary.rb:62:in `initialize'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database/dictionary.rb:32:in `new'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database/dictionary.rb:32:in `block in entries'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database/dictionary.rb:31:in `map'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database/dictionary.rb:31:in `entries'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database/gitlab_schema_info.rb:123:in `all_table_allows'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database/gitlab_schema_info.rb:99:in `add_table_specific_allows'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database/gitlab_schema_info.rb:18:in `initialize'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database/gitlab_schema_info.rb:28:in `new'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database/gitlab_schema_info.rb:28:in `load_file'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database.rb:81:in `block in all_gitlab_schemas'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database.rb:81:in `map'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database.rb:81:in `all_gitlab_schemas'
/Users/tkuah/code/gdk-ee/gitlab/lib/gitlab/database.rb:122:in `schemas_to_base_models'
[...]

Issue #471, opened 2024-01-30
1.18.x: `compile_cache/iseq.rb:64:in 'fetch': Invalid argument - (Errno::EINVAL)`


casperisfine:
I started seeing this error after the upgrade, somehow it wasn't caught during my integration testing while developing `1.18`.

Looking into it.

casperisfine:
Example corrupted header:

```
"\x05\x00\x00\x00\x881R9\xC1\x06z\xC0/IL(\x9E\x03\x00\x00\x00\x00\x00\x00\xDA\x16\xB9e\x00\x00\x00\x00\x00\x00\x00\x00\xE8\x7F\x00\x00\xB8\xBF0\xE4\xD4\e\xB4\xB2\x01_C\x01\x00\x00\x00\x00\xF8\x17z\x8B\xE8\x7F\x00\x00"
```

And the whole entry was `2376` bytes.

PR #476, opened 2024-01-31
Fix a cache corruption issue during revalidation

casperisfine:
Fix: https://github.com/Shopify/bootsnap/issues/474
Fix: https://github.com/Shopify/bootsnap/issues/471

As noticed by Stan Hu, upon revalidation, we'd corrupt the `data_size` field, as it wasn't initialized at all on the cache key we computed from the source file.

The only thing we actually want to update is the `mtime`.

Co-Authored-By: @stanhu (we didn't pair, but spotting the root cause was 99% of the work, so...)

For now I leave the revalidation as optional, but will likely flip it back on in another release next week, once it has run for a while on our CI without issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Fix the cache corruption issue in the revalidation feature. See #474.
+
 # 1.18.2
 
 * Disable stale cache entries revalidation by default as it seems to cause cache corruption issues. See #471 and #474.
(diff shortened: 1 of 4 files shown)


dlsteuer:
I had this error pop up yesterday and it's still happening to me even on 1.18.3, are there any workarounds?

casperisfine:
@dlsteuer have you cleared your cache? To be fair I should have bumped the cache version, my bad.

casperisfine:
Next release will. https://github.com/Shopify/bootsnap/commit/811b34adb8a404c6fa8017f6c6416b03eee74a66

dlsteuer:
> @dlsteuer have you cleared your cache? To be fair I should have bumped the cache version, my bad.

Is this just a matter of deleting the cache folder?

Issue #495, opened 2024-11-26
Bootsnap causes kamal setup to hang in Ruby 3.3.5


iiewad:
I have a new Rails app,  There are some env
```
ruby 3.3.5
rails 8.0.0
kamal 2.3.0
```
Dockerfile
```Dockerfile
# Install application gems
RUN bundle install
RUN rm -rf ~/.bundle/ "${BUNDLE_PATH}"/ruby/*/cache "${BUNDLE_PATH}"/ruby/*/bundler/gems/*/.git
RUN rm -rf tmp/cache/bootsnap*
RUN bundle exec bootsnap precompile --gemfile

# Copy application code
COPY . .

# Precompile bootsnap code for faster boot times
RUN bundle exec bootsnap precompile app/ lib/

# Precompiling assets for production without requiring secret RAILS_MASTER_KEY
RUN SECRET_KEY_BASE_DUMMY=1 ./bin/rails assets:precompile
```
When I run kamal setup
```
 DEBUG [c688ef86]       #15 /bin/sh -c bundle install
 DEBUG [c688ef86]       #15 CACHED
 DEBUG [c688ef86]
 DEBUG [c688ef86]       #16 /bin/sh -c rm -rf ~/.bundle/ "${BUNDLE_PATH}"/ruby/*/cache "${BUNDLE_PATH}"/ruby/*/bundler/gems/*/.git
 DEBUG [c688ef86]       #16 CACHED
 DEBUG [c688ef86]
 DEBUG [c688ef86]       #17 /bin/sh -c rm -rf tmp/cache/bootsnap*
 DEBUG [c688ef86]       #17 CACHED
 DEBUG [c688ef86]
 DEBUG [c688ef86]       #18 /bin/sh -c bundle exec bootsnap precompile --gemfile
```

mattgibson:
Switching to Ruby 3.2.6 fixes it for me too

caiohsramos:
I have the exact same issue. My specs are:
- M2 Mac with macOS 15.2
- Docker Desktop 4.37.2
- Docker Engine v27.4.0

Tried with different Docker "Virtual Machine Options" and with a few buildkit versions for the builder. The command is
```
docker buildx build --push --platform linux/amd64 --builder kamal-local-docker-container -t ghcr.io/... -t ghcr.io/... --label service="..." --file Dockerfile .
```
[...]

PR #501, opened 2025-05-14
Improve CLI parallelism

casperisfine:
- Avoid spawning more processes than what the cgroups cpu quota allows.
- Try to detect QEMU fork bug.

Fix: https://github.com/Shopify/bootsnap/issues/495
Closes: https://github.com/Shopify/bootsnap/pull/497

I'm waiting on a confirmation that the bug detection work as expected, once I have it I'll cut a release.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
+          cache-version: 2
       - run: bundle exec rake
 
   rubocop:
@@ -36,6 +37,7 @@ jobs:
         with:
           ruby-version: '3.3'
           bundler-cache: true
+          cache-version: 2
       - run: bundle exec rubocop
 
   rubies:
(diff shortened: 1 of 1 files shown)


lib/bootsnap/cli/worker_pool.rb
```diff
@@ -1,16 +1,88 @@
...
+
+        def cpu_quota
+          if RbConfig::CONFIG["target_os"].include?("linux")
+            if File.exist?("/sys/fs/cgroup/cpu.max")
+              # cgroups v2: https://docs.kernel.org/admin-guide/cgroup-v2.html#cpu-interface-files
+              cpu_max = File.read("/sys/fs/cgroup/cpu.max")
+              return nil if cpu_max.start_with?("max ") # no limit
+
+              max, period = cpu_max.split.map(&:to_f)
+              max / period
```
n00dle, line 45: These lines (and lines 53/54) are returning a float, causing issues later on when trying to spawn the worker pool. This is the error from our CI:

```
/home/circleci/tmp/vendor/bundle/ruby/3.1.0/gems/bootsnap-1.18.5/lib/bootsnap/cli/worker_pool.rb:164:in `spawn': undefined method `times' for 4.0:Float (NoMethodError)

        @workers = @size.times.map { Worker.new(@jobs) }
                        ^^^^^^
```

casperisfine: Woops.

PR #523, opened 2026-01-15
Properly fix opendir failing with `errno = 0`

byroot:
Thanks to the person who put me on the right track on Reddit.

I believe the bug happens when one of the `rb_ary_new*` triggers a GC, which may invoke code that may reset `errno` (typically a file finalizer).


ngan:
@byroot will you be releasing a new version with this fix?

PR #528, opened 2026-02-03
Fix bundle path check and consolidate common code in PathScanner

Umofomia:
## Summary

Thanks for merging #526! While reviewing that fix and `PathScanner` in general, I noticed other behavioral differences between `ruby_call` and `native_call`, so I made the following fixes:

- Fix bug where `native_call` compared relative paths against absolute `BUNDLE_PATH` in nested directory scanning
- Extract common setup code from `ruby_call` and `native_call` into a shared `prepare_scan` helper method to minimize potential behavioral differences between the two
- Optimize `walk` to receive pre-partitioned ignored directories instead of recomputing on each recursive call
- Use consistent variable naming in both `ruby_call` and `native_call` so it's easier to compare the two implementations

## Problem

I noticed `native_call` had a bug in how it checks against `BUNDLE_PATH` here:
https://github.com/rails/bootsnap/blob/c2ef9a36c3ae67338e57aacd98d3399333ea0bc2/lib/bootsnap/load_path_cache/path_scanner.rb#L86-L91

After `dirs.map!`, `dir` is a relative path but `BUNDLE_PATH` is an absolute path, so the comparison will never match. In practice, the code often works because it skips entries starting with `.`, so `.bundle` will generally be skipped, but it would fail if `BUNDLE_PATH` is set to something like `/path/to/vendor/bundle` instead, which doesn't have a dot prefix.

This change fixes the code in `native_call` to compare against the absolute path instead and added a test that captures the case that would have failed in the original code.

### DIFF
--- a/test/load_path_cache/path_scanner_test.rb
+++ b/test/load_path_cache/path_scanner_test.rb
@@ -49,56 +49,69 @@ module Bootsnap
       end
 
       def test_ignores_directories_by_name
-        original_ignored = PathScanner.ignored_directories
-        PathScanner.ignored_directories = ["ignored"]
+        with_ignored_directories(["ignored"]) do
+          Dir.mktmpdir do |dir|
+            FileUtils.mkdir_p("#{dir}/ignored")
+            FileUtils.mkdir_p("#{dir}/included")
+            FileUtils.touch("#{dir}/ignored/a.rb")
+            FileUtils.touch("#{dir}/included/b.rb")
 
-        Dir.mktmpdir do |dir|
-          FileUtils.mkdir_p("#{dir}/ignored")
-          FileUtils.mkdir_p("#{dir}/included")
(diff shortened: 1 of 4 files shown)


test/load_path_cache/path_scanner_test.rb
```diff
@@ -100,6 +100,31 @@ def test_ignores_nested_directories_by_absolute_path
       ensure
         PathScanner.ignored_directories = original_ignored
       end
+
+      def test_excludes_bundle_path_in_nested_directories
+        original_bundle_path = PathScanner::BUNDLE_PATH
+
+        Dir.mktmpdir do |dir|
+          bundle_path = "#{dir}/vendor/bundle/"
+          PathScanner.instance_eval do
```
byroot: You don't need that instance_eval.

Look for `stub_const` in `test/load_path_cache/store_test.rb`. It can be made shared and used it.

PR #535, opened 2026-04-26
Allow to substitute the Ruby compiler

byroot:
If `Bootsnap::CompileCache::ISeq.compiler_selector` is set to a proc, it will be called for every loaded Ruby file, allowing to intercept or replace the compiler.

Bootsnap provides 3 default compilers:

  - `Bootsnap::CompileCache::ISeq::DEFAULT`
  - `Bootsnap::CompileCache::ISeq::FROZEN_STRING_LITERAL`
  - `Bootsnap::CompileCache::ISeq::MUTABLE_STRING_LITERAL`

Other gems or private code can define and write custom compilers, however note that there is no guarantee the interface will remain stable (even though it historically almost never changed).

### DIFF
--- a/lib/bootsnap/compile_cache/iseq.rb
+++ b/lib/bootsnap/compile_cache/iseq.rb
@@ -7,7 +7,8 @@ module Bootsnap
   module CompileCache
     module ISeq
       class << self
-        attr_reader(:cache_dir)
+        attr_reader :cache_dir
+        attr_accessor :compiler_selector, :default_compiler
 
         def cache_dir=(cache_dir)
           @cache_dir = cache_dir.end_with?("/") ? "#{cache_dir}iseq" : "#{cache_dir}-iseq"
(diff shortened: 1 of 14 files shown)


byroot:
Interestingly, it seems like there's a regression in ruby-head, will look into it.

byroot:
Seems like it's already fixed upstream, setup-ruby must just be lagging behind a little bit: https://github.com/ruby/ruby/pull/16789

lib/bootsnap/compile_cache/iseq.rb
palkan, line 102: @byroot Looks like we have a potential problem here: if Bootsnap is not enabled when we have Coverage enabled then the new _compiler_ feature is not used either. In other words, literals in tests (assuming the Coverage is on) won't be frozen. So, prod/dev and test envs won't match. (I've just discovered that coverage issue lately and have a "fix" [here](https://github.com/ruby-next/require-hooks/commit/af90ceead6f9105c989e9be29723b5dd16fe1da6) which is probably too much for Bootsnap).

byroot: > which is probably too much for Bootsnap

Not sure, I'll have to check. Historically Bootsnap just bails out if coverage is enabled: https://github.com/rails/bootsnap/blob/dc57c2069a38710d30077d988229c4d6f7f4292d/lib/bootsnap/compile_cache/iseq.rb#L102

I'll have to research a bit, but I suspect we can call `input_to_output` instead of bailing out. The only important part is that we don't cache ISeq generated while coverage is on and that we don't load iseq from cache either.

But I suspect we can call `.compile_file` anyways.

byroot: Here's what I have in mind: https://github.com/rails/bootsnap/pull/536

Not 100% sure it will work, but it might?

palkan: > But I suspect we can call .compile_file anyways.

I'm afraid that won't help: `.compile_file` (and other iseq methods) doesn't _activate_ coverage tracking. There is no `iseq_setup_coverage` call made when we eval iseqs (it's only present in regular `require` code paths). I tried to find a way to overcome this but only made it work with an eval hack (and enabling coverage for eval) 🤷‍♂️. There are some tests in the `require-hooks` repo you can play with.

P.S. We found this issue in one of the apps recently where we've integrated Freezolite: the coverage was lost.

byroot: I see. I now have a minimal repro thanks to your tests.

I'll see if perhaps we can fix coverage to work with `ISeq.compile`, but even if we can fix it in Ruby, that will remain a big issue for a while.

I'm not sure what we can do about this. There's of course the options of assigning `.compile_option` like freezolite does, but that means another entirely different codepath :/

palkan: > ven if we can fix it in Ruby, that will remain a big issue for a while.

Yeah( I think, it worths looking at this problem from the Ruby side. Either _activate_ coverage for `.compile_path`, `iseq.eval` and such (I wonder why we don't have and whether it's intentional), or enhance Coverage API to allow registering paths from Ruby (or at least from C via extensions—also not possible today). If manage to have smth like that, in a few years we will be good (that could even happen before frozen literals by default will ship 😁).

> I'm not sure what we can do about this. There's of course the options of assigning .compile_option like freezolite does, but that means another entirely different codepath :/

Right. As far as I researched, currently there are two workarounds for selective literals freezing: 1) `.compile_option` + `Kernel#require` patch; 2) `.compile_path` / `iseq.eval` + enabling eval coverage + faking Coverage registration via `eval("<empty lines>", path)` (that's what I have in require-hooks@master right now; we tested it, works well).

byroot: > I think, it worths looking at this problem from the Ruby side

I think this will do: https://github.com/ruby/ruby/pull/16805

palkan: Awesome! Thanks!

Do you plan to revert the frozen compiler activation for current/older Rubies?

byroot: I'm not yet decided. I might try to work around it, if the workaround is only used when coverage is enabled, I think it's acceptable.

palkan: Awesome.
Please, let me know if you find a better workaround then pre-evaling _nothing_ for coverage registration.

(I'm talking about this:)

```ruby
# hello.rb
str = "Hello"
puts str + " (#{frozen?(str)})"

# irb>
require "coverage"
Coverage.start(lines: true, eval: true) # eval is required

path = File.join(Dir.pwd, "hello.rb")

eval("\n" * File.read(path).lines.size, TOPLEVEL_BINDING, path, 1)
RubyVM::InstructionSequence.compile_file(path).eval

Coverage.peek_result #=> { "../hello.rb" => [1, 1], ... } 
```

PR #536, opened 2026-04-28
Namespace bytecode compiled with coverage on

byroot:


scottjacobsen:
I'm looking at patching `fetch` like this - to load everything but app code from the bootsnap cache - since we don't care about gem coverage:

```ruby
if (ENV["COVERAGE"] || ENV["COVERAGE_UPDATE"]) && defined?(Bootsnap::CompileCache::ISeq)
  module GemIseqCacheUnderCoverage
    GEM_ROOTS = [Bundler.bundle_path.to_s, *Gem.path, Gem.default_dir, RbConfig::CONFIG["rubylibdir"]].uniq.freeze

    def fetch(path, cache_dir: Bootsnap::CompileCache::ISeq.cache_dir)
      return super unless coverage_on? && GEM_ROOTS.any? { |root| path.start_with?(root) }

      compiler = Bootsnap::CompileCache::ISeq::DEFAULT
      Bootsnap::CompileCache::Native.fetch(cache_dir, compiler.namespace, path, compiler, nil)
    rescue RuntimeError
      nil
    end
  end

  Bootsnap::CompileCache::ISeq.singleton_class.prepend(GemIseqCacheUnderCoverage)
  ```
  
  Curious about the approach in this closed PR though - is there another stream of related working going on?

byroot:
See https://github.com/rails/bootsnap/pull/535

PR #546, opened 2026-05-12
Add rubocop performance

davidenglishmusic:
This pr adds rubocop-performance and fixes a few infractions. The performance gains are not big since they are in startup and test but I think still worthwhile.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,6 @@
+plugins:
+  - rubocop-performance
+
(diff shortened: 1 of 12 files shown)


byroot:
None of these make a measurable difference. But overall I think they’re positive changes by using simpler API so fine by me.

byroot:
Nb: the failing cops are my fault I’ll fix it in main

PR #559, opened 2026-07-22
Remove `ruby2_keyword` usage

Earlopain:
Also a bit easier to understand.


byroot:
Not too sure how I feel about this. This path is quite tricky to get right across 2.x and 3.x, so IMO `ruby2_keywords` kinda make it easier.

If we were to drop 2.x then yes maybe, but I'm quite conservative on dropping old version support.

Earlopain:
Ah, my bad. I thought it was 3.0 already. Wanted to demonstrate for https://bugs.ruby-lang.org/issues/22205. Should have choosen something different 😅

PR #560, opened 2026-08-05
Replace Date Marshalling with recursive msgpack

skipkayhil:
It's more compact, faster, and doesn't use Marshal.

```ruby
# frozen_string_literal: true

require "date"
require "msgpack"
require "benchmark/ips"

def codec(**opts)
  f = MessagePack::Factory.new
  f.register_type(0x01, Date, **opts)
  f
end

CODECS = {
  "current (Marshal)" => codec(
    packer: ->(d) { Marshal.dump(d) },
    unpacker: ->(p) { Marshal.load(p) },
  ),

  "3 ints" => codec(
    packer: ->(d, pk) { pk.write(d.year); pk.write(d.month); pk.write(d.day) },
    unpacker: ->(u) { Date.new(u.read, u.read, u.read) },
    recursive: true,
  ),

  # similar to paquito, but uses unpack1/shifting to avoid array allocation
  "bitpack" => codec(
    packer: ->(d) { [(d.year << 9) | (d.month << 5) | d.day].pack("l<") },
    unpacker: ->(p) { v = p.unpack1("l<"); Date.new(v >> 9, (v >> 5) & 0xF, v & 0x1F) },
  ),
}

DATE = Date.new(2024, 6, 15)

CODECS.each { |n, f| raise n unless f.load(f.dump(DATE)) == DATE }

Benchmark.ips do |x|
  CODECS.each do |n, f|
    b = f.dump(DATE)
    puts " %-22s %2d bytes  %s" % [n, b.bytesize, b.bytes.map { |x| "%02X" % x }.join(" ")]

    b = f.dump(10.times.map { DATE })
    x.report(n) { f.load(b) }
  end
  x.compare!
end
```

```
$ ruby bench.rb
 current (Marshal)      36 bytes  C7 21 01 04 08 55 3A 09 44 61 74 65 5B 0B 69 00 69 03 3D 8B 25 69 00 69 00 69 00 66 0C 32 32 39 39 31 36 31
 3 ints                  8 bytes  C7 05 01 CD 07 E8 06 0F
 bitpack                 6 bytes  D6 01 CF D0 0F 00
ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [x86_64-linux]
Warming up --------------------------------------
   current (Marshal)     2.616k i/100ms
              3 ints    16.045k i/100ms
             bitpack    11.230k i/100ms
Calculating -------------------------------------
   current (Marshal)     27.840k (± 3.9%) i/s   (35.92 μs/i) -    141.264k in   5.074188s
              3 ints    157.179k (± 2.4%) i/s    (6.36 μs/i) -    786.205k in   5.001970s
             bitpack    110.688k (± 2.5%) i/s    (9.03 μs/i) -    561.500k in   5.072816s

Comparison:
           3 ints:   157179.1 i/s
          bitpack:   110688.0 i/s - 1.42x  slower
current (Marshal):    27839.7 i/s - 5.65x  slower
```

I included an approach similar to Paquito ("bitpack") in the benchmark, but it uses unpack1 and shifting to avoid allocating an array when unpacking (Paquito's implementation uses unpack so it allocates). Like Paquito's implementation, it only uses 6 bytes but is slower than the "3 ints" approach.

I went with "3 ints" because its the fastest at unpacking, the implementation is very simple, and it only uses an extra 2 bytes per Date. Compared to the 36 bytes currently used by Marshal, the extra 2 seems negligible.

One advantage of Marshal.dump is that it automatically handles subclasses of Date, so I also added a "simple" DateTime type mirroring Paquito.

### DIFF
--- a/lib/bootsnap/compile_cache/yaml.rb
+++ b/lib/bootsnap/compile_cache/yaml.rb
@@ -89,16 +89,58 @@ module Bootsnap
             unpacker: MessagePack::Time::Unpacker,
           )
 
-          marshal_fallback = {
+          factory.register_type(
+            0x01,
+            Date,
+            packer: lambda { |date, packer|
+              packer.write(date.year)
+              packer.write(date.month)
+              packer.write(date.day)
+            },
+            unpacker: lambda { |unpacker|
+              ::Date.new(unpacker.read, unpacker.read, unpacker.read)
+            },
+            recursive: true,
+          )
+
+          factory.register_type(
+            0x02,
+            Regexp,
             packer: ->(value) { Marshal.dump(value) },
             unpacker: ->(payload) { Marshal.load(payload) },
-          }
-          {
-            Date => 0x01,
-            Regexp => 0x02,
-          }.each do |type, code|
-            factory.register_type(code, type, marshal_fallback)
-          end
+          )
+
+          factory.register_type(
+            0x03,
+            DateTime,
+            packer: lambda { |dt, packer|
+              packer.write(dt.year)
+              packer.write(dt.month)
+              packer.write(dt.day)
+              packer.write(dt.hour)
+              packer.write(dt.minute)
+
+              sec = dt.sec + dt.sec_fraction
+              packer.write(sec.numerator)
+              packer.write(sec.denominator)
+
+              offset = dt.offset
+              packer.write(offset.numerator)
+              packer.write(offset.denominator)
+            },
+            unpacker: lambda { |unpacker|
+              ::DateTime.new(
+                unpacker.read, # year
+                unpacker.read, # month
+                unpacker.read, # day
+                unpacker.read, # hour
+                unpacker.read, # minute
+                Rational(unpacker.read, unpacker.read), # sec fraction
+                Rational(unpacker.read, unpacker.read), # offset fraction
+              )
+            },
+            recursive: true,
+          )
 
           self.msgpack_factory = factory
 
--- a/lib/bootsnap/compile_cache/yaml.rb
+++ b/lib/bootsnap/compile_cache/yaml.rb
@@ -142,6 +142,15 @@ module Bootsnap
             recursive: true,
           )
 
+          require "msgpack/bigint"
+          factory.register_type(
+            0x04,
+            Integer,
+            packer: MessagePack::Bigint.method(:to_msgpack_ext),
+            unpacker: MessagePack::Bigint.method(:from_msgpack_ext),
+            oversized_integer_extension: true,
+          )
+
           self.msgpack_factory = factory
 
           self.supported_options = []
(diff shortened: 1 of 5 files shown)


lib/bootsnap/compile_cache/yaml.rb
```diff
@@ -97,16 +97,58 @@ def visit(target)
...
-            }.each do |type, code|
-              factory.register_type(code, type, marshal_fallback)
-            end
+            )
+
+            factory.register_type(
+              0x03,
+              DateTime,
+              packer: lambda { |dt, packer|
+                packer.write(dt.year)
```
byroot: Years can be arbitrary large, so we need `Msgpack::Bigint`.

skipkayhil: 👍 pushed up another commit to add bigint. Interestingly though, Psych currently won't load a Date that has a year with more or less than 4 digits: https://github.com/ruby/psych/blob/8aaf6c21efb9ddeab0c3fb8cd8064525543e87ef/lib/psych/scalar_scanner.rb#L76

lib/bootsnap/compile_cache/yaml.rb
```diff
@@ -149,6 +149,25 @@ def visit(target)
...
+                  packer.write(MessagePack::Bigint.to_msgpack_ext(value))
+                },
+                unpacker: lambda { |unpacker|
+                  MessagePack::Bigint.from_msgpack_ext(unpacker.read)
+                },
+                oversized_integer_extension: true,
+                recursive: true,
+              )
+            rescue LoadError
+              # msgpack < 1.5.0
```
byroot: We can just require `>= 1.5.0` in the gemspec.

lib/bootsnap/compile_cache/yaml.rb
```diff
@@ -97,15 +97,76 @@ def visit(target)
...
+            factory.register_type(
+              0x01,
+              Date,
+              packer: lambda { |date, packer|
+                packer.write(date.year)
+                packer.write(date.month)
+                packer.write(date.day)
+              },
+              unpacker: lambda { |unpacker|
+                ::Date.new(unpacker.read, unpacker.read, unpacker.read)
```
skipkayhil: Another edge case we may want to consider: https://github.com/ruby/psych/blob/8aaf6c21efb9ddeab0c3fb8cd8064525543e87ef/lib/psych/scalar_scanner.rb#L78

Psych >= 5 loads dates as `GREGORIAN`, so we may need to do that as well to ensure things round trip correctly?

lib/bootsnap/compile_cache/yaml.rb
```diff
@@ -149,6 +149,25 @@ def visit(target)
...
+                0x04,
+                Integer,
+                packer: lambda { |value, packer|
+                  packer.write(MessagePack::Bigint.to_msgpack_ext(value))
+                },
+                unpacker: lambda { |unpacker|
+                  MessagePack::Bigint.from_msgpack_ext(unpacker.read)
+                },
+                oversized_integer_extension: true,
+                recursive: true,
```
byroot: You don't need recursion, it's wasteful, see: https://github.com/Shopify/paquito/blob/74eff7a0888e75d828ab971bb66206d5d62a4fda/lib/paquito/types.rb#L364-L366

PR #563, opened 2026-08-08
Ignore YJIT status in the compile cache key

### LINKED ISSUE #562: Compile cache is stale when YJIT is toggled because RUBY_DESCRIPTION is part of the cache key
The compile cache key incorporates `RUBY_DESCRIPTION`, which changes when YJIT is enabled:

```
$ RUBYOPT="--yjit" ruby -e "puts RUBY_DESCRIPTION"
ruby 3.3.11 (2026-03-26 revision 1f2d15125a) +YJIT [x86_64-linux]
$ ruby -e "puts RUBY_DESCRIPTION"
ruby 3.3.11 (2026-03-26 revision 1f2d15125a) [x86_64-linux]
```

We recently ran into a problem where we belatedly realized that the Bootsnap cache was being marked `stale` because YJIT was enabled in production, but it was not enabled during the `bootsnap precompile` phase. As a result, we saw a 0% hit rate and full ISeq recompilation on every boot. It also affects `RubyVM::YJIT.enable` (which Rails 7.2 invokes) at runtime, which also mutates `RUBY_DESCRIPTION`.

I think this is a false positive. The compile cache stores serialized `ISeq` (`RubyVM::InstructionSequence.to_binary`). YJIT compiles machine code from the ISeq at runtime; it does not change the ISeq binary format/contents, nor `RubyVM::InstructionSequence.compile_option`. A binary ISeq produced with YJIT off loads and runs correctly with YJIT on and vice versa, so YJIT status has no bearing on cache validity.

Proposal: Strip the ` +YJIT ` marker out of `RUBY_DESCRIPTION` before hashing it in `get_ruby_version_digest()`.

stanhu:
The compile cache key includes a digest of `RUBY_DESCRIPTION`, which changes when YJIT is enabled (it gains a ` +YJIT` marker). Because YJIT is a runtime JIT that does not affect the serialized instruction sequences that are cached, this caused the entire compile cache to be treated as stale and recompiled whenever YJIT was enabled at runtime (via `--yjit`, `RUBYOPT`, or `RubyVM::YJIT.enable`) but not at precompile time, or vice versa. A binary ISeq compiled with YJIT off loads and runs correctly with YJIT on and vice versa, so this was purely a false- positive invalidation.

Strip the YJIT marker from `RUBY_DESCRIPTION` before hashing it so the digest is stable regardless of YJIT. On release builds the marker is ` +YJIT`; builds with `YJIT_SUPPORT` defined append a token such as ` +YJIT dev` (see Ruby's `YJIT_DESCRIPTION`), so an optional trailing ` <token>` is also consumed. Other markers such as ` +PRISM` are left intact.

The added subprocess test populates the cache with YJIT disabled and then loads the same file with `RUBYOPT=--yjit`, asserting the result is a cache hit rather than stale.

Closes https://github.com/rails/bootsnap/issues/562

### DIFF
--- a/ext/bootsnap/bootsnap.c
+++ b/ext/bootsnap/bootsnap.c
@@ -384,7 +384,7 @@ bs_compile_option_crc32_set(VALUE self, VALUE crc32_v)
 static uint64_t
 get_ruby_version_digest(void)
 {
-  uint64_t hash = fnv1a_64_str(rb_const_get(rb_cObject, rb_intern("RUBY_DESCRIPTION")));
+  uint64_t hash = fnv1a_64_str(rb_const_get(rb_mBootsnap, rb_intern("RUBY_CACHE_KEY")));
   hash = fnv1a_64_iter(hash, (unsigned char *)&bootsnap_cache_version, sizeof(bootsnap_cache_version));
   return hash;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+* Don't invalidate the compile cache when YJIT is toggled. YJIT is a runtime JIT and doesn't change the
+  serialized instruction sequences that are cached, but enabling it (via `--yjit`, `RUBYOPT`, or
+  `RubyVM::YJIT.enable`) adds a ` +YJIT` marker to `RUBY_DESCRIPTION` (` +YJIT <token>` on `YJIT_SUPPORT`
+  builds), which is part of the cache key. This previously discarded the entire compile cache whenever YJIT
+  was enabled at runtime but not at precompile time (or vice versa). The marker is now stripped before hashing.
+
 * Fix `CompileCache::Native.fetch` and `.precompile` reading a non-`String` path argument (e.g. a `Pathname`)
   with `RSTRING_PTR`. Regression from 1.24.0.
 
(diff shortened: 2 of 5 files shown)


ext/bootsnap/bootsnap.c
```diff
@@ -384,7 +384,27 @@ bs_compile_option_crc32_set(VALUE self, VALUE crc32_v)
...
+   *
+   * On release builds the marker is " +YJIT"; builds with YJIT_SUPPORT defined
+   * append a token, e.g. " +YJIT dev" (see Ruby's YJIT_DESCRIPTION), so we also
+   * consume an optional trailing " <token>". Other markers such as " +PRISM"
+   * are left intact.
+   */
+  static const char yjit_pattern[] = " \\+YJIT( \\w+)?";
+  VALUE yjit_regexp = rb_reg_new(yjit_pattern, sizeof(yjit_pattern) - 1, 0);
+  ruby_description = rb_funcall(ruby_description, rb_intern("gsub"), 2,
+				yjit_regexp, rb_str_new_cstr(""));
```
byroot: That's the sort of thing we should just do in ruby and pass to the C extension during init.

stanhu: Makes sense, thanks for taking care of that.

